PR #323, opened 2019-10-10
[FIX] qweb: event modifier with arg in t-foreach

### LINKED ISSUE #266: event modifiers, arguments,  t-foreach: wrong result
```xml
<t t-foreach="projects" t-as="project">
<a href="#" t-on-click.prevent="editProject(project.id)">Éditer</a>
</t>
```

With the `prevent` modifier, this will result in all event handlers bound to the last row. If we remove the `prevent` modifier and we move the `preventDefault` to the `editProject` method, it works as expected

aab-odoo:
Closes #266

### DIFF
--- a/src/qweb/extensions.ts
+++ b/src/qweb/extensions.ts
@@ -49,25 +49,35 @@ QWeb.addDirective({
     );
     ctx.closeIf();
     let params = extraArgs ? `owner, ${ctx.formatExpression(extraArgs)}` : "owner";
-    let handler;
     if (mods.length > 0) {
-      handler = `function (e) {`;
+      let handler = `function (e) {`;
       handler += mods
         .map(function(mod) {
           return MODS_CODE[mod];
         })
         .join("");
-      handler += `context['${handlerName}'].call(${params}, e);}`;
+      if (!extraArgs) {
+        handler += `context['${handlerName}'].call(${params}, e);}`;
+        ctx.addLine(
+          `extra.handlers['${eventName}' + ${nodeID}] = extra.handlers['${eventName}' + ${nodeID}] || ${handler};`
+        );
+        ctx.addLine(`p${nodeID}.on['${eventName}'] = extra.handlers['${eventName}' + ${nodeID}];`);
+      } else {
+        const handlerKey = `handler${ctx.generateID()}`;
+        ctx.addLine(`const ${handlerKey} = context['${handlerName}'].bind(${params});`);
+        handler += `${handlerKey}(e);}`;
+        ctx.addLine(`p${nodeID}.on['${eventName}'] = ${handler};`);
+      }
     } else {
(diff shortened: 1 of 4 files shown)


src/qweb/extensions.ts
```diff
@@ -51,13 +51,15 @@ QWeb.addDirective({
...
+      const handlerKey = `handler${ctx.generateID()}`;
+      ctx.addLine(`const ${handlerKey} = context['${handlerName}'].bind(${params});`);
       handler = `function (e) {`;
       handler += mods
         .map(function(mod) {
           return MODS_CODE[mod];
         })
         .join("");
-      handler += `context['${handlerName}'].call(${params}, e);}`;
+      handler += `${handlerKey}.call(owner, e);}`;
```
ged-odoo: if it is already bound, maybe you can simply do that: 
```suggestion
      handler += `${handlerKey}(e);}`;
```
not sure if it works though

tests/qweb/qweb.test.ts
```diff
@@ -1086,6 +1086,37 @@ describe("t-on", () => {
...
+      projects: [{id: 1, name: 'Project 1'}, {id: 2, name: 'Project 2'}],
+
+      onEdit(projectId, ev) {
+        expect(ev.defaultPrevented).toBe(true);
+        steps.push(projectId);
+      }
+    }
+
+    const node = renderToDOM(qweb, "test", owner, { handlers: [] });
+
```
ged-odoo: you should probably assert that the result of the rendering is what we expect. something like
```js
  expect(node.outerHTML).toBe("<div><a href="#">Edit Project 1</a><a href="#">Edit Project 2</a></div>");
```

aab-odoo:
@ged-odoo done

tests/qweb/__snapshots__/qweb.test.ts.snap
```diff
@@ -1778,12 +1825,13 @@ exports[`t-on t-on with self and prevent modifiers (order matters) 1`] = `
     if (!context['onClick']) {
         throw new Error('Missing handler \\\\'' + 'onClick' + \`\\\\' when evaluating template 'test'\`)
     }
-    extra.handlers['click' + 2] = extra.handlers['click' + 2] || function (e) {if (e.target !== this.elm) {return}e.preventDefault();context['onClick'].call(owner, e);};
+    const handler3 = context['onClick'].bind(owner);
```
ged-odoo: hmm, it's annoying now that I see this: we create the binding every render, even though we only need it when it is not already present in the handlers object.

Would it not be better if the generated code would be something like this:
```js
  if (!extra.handlers['click' + 2]) {
   const handler3 = context['onClickPrevented'].bind(owner);
    extra.handlers['click' + 2] = function (e) {e.preventDefault();handler3(e);};
}
```

PR #349, opened 2019-10-17
[FIX] playground: update samples to make them work on firefox

### LINKED ISSUE #335: playground broken
Go to https://odoo.github.io/owl/playground/ and click run:

Result: SyntaxError: bad method definition

ged-odoo:
Sadly, Firefox does not support yet static class fields, so we need to
use an equivalent, but slightly not as nice, syntax.

closes #335


ged-odoo:
@aab-odoo when you approve a PR, you can merge it immediately :)

PR #383, opened 2019-10-22
[FIX] store: getter as 3rd argument of useStore

alexkuhn:
Revision on 895fe7c60fa6d84e4e95d09e9d5e229a1103548c

Commit above turned old static method `mapStoreToProps`
into the hook `useStore`. For some reasons, it removed
`getters` as the 3rd argument of the selector, which this
commit re-introduced.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -84,7 +84,7 @@ const isStrictEqual = (a, b) => a === b;
 export function useStore(selector, options: SelectorOptions = {}): any {
   const component: Component<any, any> = Component.current!;
   const store = options.store || (component.env.store as Store);
-  let result = selector(store.state, component.props);
+  let result = selector(store.state, component.props, store.getters);
   const hashFn = store.observer.deepRevNumber.bind(store.observer);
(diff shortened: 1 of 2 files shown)


ged-odoo:
there is a `useGetters` hook now, that is why it is not available in the `useStore` hook.  But now that you mention it, it is not well documented...

ged-odoo:
I assume that the `useGetters` is good enough for you?

alexkuhn:
Yes, I should have seen this hook exists :)

alexkuhn:
Just wondering if we really need `props` as 2nd argument of `useStore`, since we could simply use `this.props` for that. It made sense when `mapStoreToProps` was static, but that's no longer the case.

alexkuhn:
I see, `useStore`'s selector is called onWillUpdateProps, which is just before `this.props` are updated with `nextProps`:
https://github.com/odoo/owl/blob/master/src/component/component.ts#L519

This looks important to mention in the doc.

PR #395, opened 2019-10-25
Issue 377 empty handler

aab-odoo:


ged-odoo:
looking really cool. could you add some documentation in the event handling section? to explain what can be done with the event handler (method name, method name with args, inline statement).

PR #408, opened 2019-10-26
[IMP] playground: log url for current owl commit

### LINKED ISSUE #405: display current commit of playground
It would be nice if the playground was linked to an exact commit instead of a revision, in order to know which intermediate fixes are already applied/ready to be tested or not.

More simply, the playground should probably be linked to a git branch (master) and updated automatically.
Maybe a parallel version could co-exist for each revisions/tags (or alternate branches if some day bug fix has to be done for several versions?).

ged-odoo:
closes #405


ged-odoo:
@seb-odoo

PR #411, opened 2019-10-27
Issue 331 t key t foreach

ged-odoo:


ged-odoo:
note: this PR is based on top of #409 

So, ideally #409 should be merged first.

PR #424, opened 2019-10-30
wip

ged-odoo:


ged-odoo:
part of #424

aab-odoo:
I guess you meant 'part of #420'

ged-odoo:
hmm, it's actually true that PR 424 is a part of PR 424...

kebeclibre:
It is a bit of a shame not to track sets, as we could expect they'd work like arrays in useState
But, I suppose implemening some kind of wrapper around add(), clear() etc... will be costly

ged-odoo:
@kebeclibre  this is one of the point of this issue.  We should find a way to do that. It involves changing the prototype to another which overrides all the state-modifying methods of Set/Map/... 

I have a few idea on how to do that, but it is certainly not high priority.  Also, this will make the observer more complex.

ged-odoo:
was implemented anyway

PR #427, opened 2019-10-30
[FIX] component: error handling: rendering with sub components

### LINKED ISSUE #425: Errors occurring in render (with sub components) are sometimes never displayed
```js
const elements = [1, 2];
const state = {1: {val: 'hello'}};
```

```xml
<t t-foreach="elements" t-as="element">
   <SubComponent p="state[element].val"/>
</t>
```

When the crash occurs (on second SubComponent), the fiber counter is 1, so the fiber can't be processed, and the error is never displayed.

aab-odoo:
Closes #425

### DIFF
--- a/src/component/fiber.ts
+++ b/src/component/fiber.ts
@@ -1,5 +1,6 @@
 import { VNode } from "../vdom/index";
 import { Component } from "./component";
+import { scheduler } from "./scheduler";
 
 /**
  * Owl Fiber Class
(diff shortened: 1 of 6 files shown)


src/component/fiber.ts
```diff
@@ -216,7 +217,9 @@ export class Fiber {
         component.catchError!(error);
       });
     } else {
+      this.root.counter = 0;
```
ged-odoo: explain

PR #448, opened 2019-11-08
[FIX] component: concurrent rendering issue

### LINKED ISSUE #421: concurrency issue: modify component between prepare and def resolved

aab-odoo:
Resolved rendering with cancelled fiber for (not yet) destroyed
component -> Cannot read property 'sel' of null

Closes #421

### DIFF
--- a/src/component/directive.ts
+++ b/src/component/directive.ts
@@ -455,7 +455,7 @@ QWeb.addDirective({
     );
     ctx.addLine(`const fiber = w${componentID}.__owl__.currentFiber;`);
     ctx.addLine(
-      `def${defID}.then(function () {if (w${componentID}.__owl__.isDestroyed) {return;} const vnode = fiber.vnode; pvnode.sel = vnode.sel; ${createHook}});`
+      `def${defID}.then(function () {if (fiber.isCancelled) {return;} const vnode = fiber.vnode; pvnode.sel = vnode.sel; ${createHook}});`
     );
     if (registerCode) {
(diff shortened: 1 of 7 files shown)


aab-odoo:
ping @alexkuhn this is the PR fixing (hopefully) your scenario

src/component/directive.ts
```diff
@@ -455,7 +455,7 @@ QWeb.addDirective({
     );
     ctx.addLine(`const fiber = w${componentID}.__owl__.currentFiber;`);
     ctx.addLine(
-      `def${defID}.then(function () {if (w${componentID}.__owl__.isDestroyed) {return;} const vnode = fiber.vnode; pvnode.sel = vnode.sel; ${createHook}});`
+      `def${defID}.then(function () {if (w${componentID}.__owl__.isDestroyed || fiber.isCancelled) {return;} const vnode = fiber.vnode; pvnode.sel = vnode.sel; ${createHook}});`
```
ged-odoo: Suggestion:
1. remove the `w${componentID}.__owl__.isDestroyed` condition,
2. add at the end of `__destroy` in component: 
```js
    if (__owl__.currentFiber) {
      __owl__.currentFiber.isCancelled = true;
    }
```

By doing this, it looks like we skip a tiny bit of work in the `prepare` step, and we do a little bit more in the `destroy` step. I think that it is correct, and that we do not need to actually call the `cancel` method of the fiber (which does more work) at this point.

PR #449, opened 2019-11-12
Master misc

ged-odoo:


doc/reference/props.md
```diff
@@ -0,0 +1,97 @@
...
+Note that `props` is an object that only makes sense from the perspective of the
+child component.
+
+## Definition
+
+The `props` object is made of every attributes defined on the template, with the
+following exceptions:
+
+- every attribute starting with `t-` are not props (they are QWeb directives),
+- `style` and `class` attributes are exluded as well (they are applied by Owl on
```
aab-odoo: excluded

doc/reference/props.md
```diff
@@ -0,0 +1,97 @@
...
+
+```xml
+<div>
+    <ComponentA a="state.a" b="'string'"/>
+    <ComponentB t-if="state.flag" model="model"/>
+    <ComponentC style="color:red;" class="left-pane" />
+</div>
+```
+
+the `props` objects contains the following keys:
```
aab-odoo: object

doc/reference/props.md
```diff
@@ -0,0 +1,97 @@
...
+
+- for `ComponentA`: `a` and `b`,
+- for `ComponentB`: `model`,
+- for `ComponentC`: empty object
+
+## Good Practices
+
+A `props` object is a collection of values that comes from the parent. As such,
```
aab-odoo: 'that come (...) they are' or 'that comes (...) it is'

doc/reference/props.md
```diff
@@ -0,0 +1,97 @@
...
+    props.a.b = 43; // Never do that!!!
+  }
+}
+```
+
+Props should be considered readonly, from the perspective of the child component.
+If there is a need to modify them, then the request to update them should be
+sent to the parent (for example, with an event).
+
+Any value can go in a props. Strings, objects, classes, or even a callback could
```
aab-odoo: or even callbacks

Issue #441, opened 2019-11-06
Concurrency issues: trigger a render when mounting the component


aab-odoo:
Let's assume the following basic component
```js
class TestWidget extends Component<any, any> {
   static template = xml`<div><t t-esc="state.val"/></div>`;
   state = useState({ val: 1 });
}
``` 

In the following scenarios, the `mount` promise is never resolved, because its fiber is cancelled by a concurrent call to `render`.

1) **Change state and call `mount` directly**
```js
const widget = new TestWidget();
widget.state.val = 2;
await widget.mount(fixture); // working fine
widget.unmount();
widget.state.val = 3;
await widget.mount(fixture); // never resolved
```
The state change has been detected before the call to `mount` but `render` is a bit delayed and called right after `mount`. However, the rendering done by `mount` is correct since it already has the correct state, so the call to `render` could be skipped. It works fine for the first call to `mount` thanks to the call to `willStart`, which delays the moment when the fiber is flagged `isRendered`, and thus, the concurrent call to `render` is correctly skipped. 
So, even if this test doesn't show it, I think that the same issue could occur for the first call to `mount`, if we manage to trigger the call to `render` after the `willStart` promise is resolved (thus when the fiber is flagged as `isRendered`, but before the next animation frame (when the fiber is patched).

2) **Call `mount` and change state directly**
```js
const widget = new TestWidget();
const prom = widget.mount(fixture);
widget.state.val = 2;
await prom; // never resolved
widget.unmount();
prom = widget.mount(fixture);
widget.state.val = 3;
await prom; // never resolved;
```
Here, the rendering done by `mount` is obsolete since the state changed afterwards. However, we would like its returned promise to resolve with the one returned by the call to `render`. Also, I think that the `mount` hooks should be called, not the `render` (patch) ones.

PR #450, opened 2019-11-13
Render called while mounting -> mount promise never resolves

aab-odoo:


tests/component/component.test.ts
```diff
@@ -3363,6 +3362,64 @@ describe("async rendering", () => {
...
+      def.resolve();
+    }, 20);
+    child.val = 5;
+    await child.render();
+    expect(fixture.innerHTML).toBe("<div><span>2|5</span></div>");
+  });
+
+  test("change state and call manually render: no unnecessary rendering", async () => {
+    // when the state is changed, the component isn't notified directly (we wait
+    // for a microtask tick before calling 'render'), so it may happen that
```
ged-odoo: it is no longer true!

src/context.ts
```diff
@@ -66,29 +66,40 @@ export class Context extends EventBus {
...
-      // components is done (if they have no children).  This is why we manually
-      // flush the scheduler.  This will force the scheduler to check
-      // immediately if they are done, which will cause their rendering
-      // promise to resolve earlier, which means that there is a chance of
-      // processing the next group in the same frame.
-      scheduler.flush();
-      await proms;
-    }
+  __notifyComponents() {
+    let rev = this.rev;
```
ged-odoo: maybe this should be done in constructor:
```js
    this.observer.notifyCB = () => {
      let rev = this.rev;
      return Promise.resolve().then(() => {
        if (rev === this.rev) {
           this.__notifyComponent();
        }
   };

```

src/component/scheduler.ts
```diff
@@ -25,12 +25,26 @@ export class Scheduler {
...
+        fiber,
+        callback: () => {
+          if (fiber.error) {
+            reject(fiber.error);
+            return;
+          }
+          resolve();
+        }
+      });
+      if (this.isRunning) {
```
ged-odoo:
```
if (!this.isRunning) {
  this.scheduleTask();
}
```

src/component/scheduler.ts
```diff
@@ -25,12 +25,25 @@ export class Scheduler {
...
+  addFiber(fiber): Promise<void> {
+    return new Promise((resolve, reject) => {
+      if (fiber.error) {
+        return reject(fiber.error);
+      }
+      this.tasks.push({
+        fiber,
+        callback: () => {
+          if (fiber.error) {
+            reject(fiber.error);
```
ged-odoo: maybe return reject(fiber.error);, just like a few lines above

src/component/fiber.ts
```diff
@@ -132,7 +161,28 @@ export class Fiber {
...
+   */
+  complete() {
+    const component = this.component;
+    if (this.target) {
+      component.__patch(this.vnode);
+      this.target.appendChild(component.el!);
+      if (document.body.contains(this.target)) {
+        component.__callMounted();
+      }
+    } else {
```
ged-odoo: `else if {`

tests/component/component.test.ts
```diff
@@ -4920,6 +4973,68 @@ describe("unmounting and remounting", () => {
...
+    expect(TestWidget.prototype.__render).toHaveBeenCalledTimes(1);
+
+    // unmount and re-mount, as in this case, willStart won't be called, so it's
+    // slightly different
+    widget.unmount();
+    prom = widget.mount(fixture);
+    widget.state.val = 3;
+    await prom;
+    expect(fixture.innerHTML).toBe("<div>3</div>");
+    expect(TestWidget.prototype.__render).toHaveBeenCalledTimes(3);
```
ged-odoo: is it correct that here, the `__patch` method was called 2 times? (once for each mount). If that is the case, it would be good to add the assertion

aab-odoo:
Closes #441 , for the record

PR #458, opened 2019-11-14
Issue 444 will start update issue

aab-odoo:


src/store.ts
```diff
@@ -114,7 +114,7 @@ export function useStore(selector, options: SelectorOptions = {}): any {
     updateFunctions.forEach(function(updateFn) {
```
ged-odoo: we should use a for loop here. foreach + function is much slower, and this is a hot code path

ged-odoo:
this pr looks good to me...

aab-odoo:
@ged-odoo I added a test case for the scenario we were talking about. It passes, thanks to [this](https://github.com/odoo/owl/blob/master/src/context.ts#L149) making [this test](https://github.com/odoo/owl/blob/master/src/context.ts#L142) always false. So if we can guarantee that onWillUnmount is called before the update callback, it's fine.

ged-odoo:
good. the test will guarantee that we keep this invariant in the future.

PR #459, opened 2019-11-14
[IMP] component: display better error message if render is empty

### LINKED ISSUE #446: better message error for component with <t/> template?
```js
class App extends owl.Component {}
const app = new App();
app.mount(document.body);
```
```xml
<templates>
  <t t-name="App"/>
</templates>
```

gives 
"Uncaught TypeError: Cannot read property 'sel' of undefined"

ged-odoo:
closes #446

### DIFF
--- a/src/component/component.ts
+++ b/src/component/component.ts
@@ -282,9 +282,7 @@ export class Component<T extends Env, Props extends {}> {
       return Promise.resolve();
     }
     if (!(target instanceof HTMLElement)) {
-      let message = `Component '${
-        this.constructor.name
-      }' cannot be mounted: the target is not a valid DOM node.`;
+      let message = `Component '${this.constructor.name}' cannot be mounted: the target is not a valid DOM node.`;
       message += `\nMaybe the DOM is not ready yet? (in that case, you can use owl.utils.whenReady)`;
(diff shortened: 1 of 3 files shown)


tests/component/component.test.ts
```diff
@@ -108,6 +108,21 @@ describe("basic widget properties", () => {
     );
   });
 
+  test.only("display an error message if result of rendering is empty", async () => {
```
ged-odoo: oups

aab-odoo: .only

src/component/component.ts
```diff
@@ -579,10 +577,14 @@ export class Component<T extends Env, Props extends {}> {
       __owl__.observer.allowMutations = false;
     }
     try {
-      fiber.vnode = __owl__.renderFn!(this, {
+      let vnode = __owl__.renderFn!(this, {
         handlers: __owl__.boundHandlers,
         fiber: fiber
       });
+      if (!vnode) {
+        throw new Error(`Rendering '${this.constructor.name}' did not return a vnode`);
```
aab-odoo: Isn't the vnode stuff an implementation detail? Actually, the problem here is that the component has an empty template

PR #466, opened 2019-11-15
Add support for Portal

aab-odoo:


src/misc/portal.ts
```diff
@@ -0,0 +1,48 @@
...
+ * a dialog, and where this dialog should be inserted in the DOM.
+ *
+ * The Portal component ensures that the communication between the content of
+ * the Portal and its parent properly works: events reaching the Portal are
+ * re-triggered on an empty <portal> node located in the parent's DOM.
+ */
+
+export class Portal extends Component<any, any> {
+  static template = xml`<portal><t t-slot="default"/></portal>`;
+  // TODO: props validation
```
ged-odoo: good idea: `static props={ target: {type: String} };` should do the trick

kebeclibre:
Idea: make the displaced element a "forceChild" of the anchor component.
when anchor is asked to re-render, it should ask the portal component to re-render that part
possible loop trap caveat

kebeclibre:
Probably needs #508 to be solved to work as we need it to

ged-odoo:
add test with a crash in a sub component

ged-odoo:
document.body.contains(Target)

doc/reference/misc.md
```diff
@@ -1,4 +1,9 @@
-# 🦉 Miscellaneous 🦉
+# 🦉 Miscellaneous 🦉 
+
+## Content
+
+- [AsyncRoot](#asyncroot)
```
ged-odoo: I would put portal first actually

ged-odoo: also, do not forget to add the `AsyncRoot` term to the owl content list in `doc/readme.md`

doc/reference/misc.md
```diff
@@ -21,3 +26,102 @@ only on subsequent ones (triggered by state or props changes).
...
+
+## `Portal`
+
+### Overview
+
+The component `Portal` is meant to be used as a transparent way to 'teleport' a piece
+of DOM to the node represented by its sole `target` props.
+
+This component aims at helping the implementation of the needed infrastructure
+for modals (as in bootstrap-modal).
```
ged-odoo: bootstrap-modal => `bootstrap-modal`

doc/reference/misc.md
```diff
@@ -21,3 +26,102 @@ only on subsequent ones (triggered by state or props changes).
...
+
+### Usage
+
+The content it will teleport is defined within the `<Portal>` node and
+internally uses the `default` [Slot](component.md#slots).
+
+This slot only contains **one** node, which in turn can have as many children as necessary.
```
ged-odoo: This slot should only ...

doc/reference/misc.md
```diff
@@ -21,3 +26,102 @@ only on subsequent ones (triggered by state or props changes).
...
+The element under which the content will be teleported is represented as a selector
+by the `target` props which only accepts a string as its value.
+
+The `target` props only supports static selector, and is not meant to be passed to `Portal`
+as a variable. Namely, `<Portal target="'body'" />` is the intended use.
+By contrast, `<Portal target="state.target" />` is not supported.
+
+The component `Portal` has no particular state, rather it is meant to be a slave to its parent,
+and ultimately just a way for the parent to teleport a piece of its own DOM elsewhere.
+
```
ged-odoo: you should add the info that there is a `<portal>` tag in the html at the original place for the portal

doc/reference/misc.md
```diff
@@ -21,3 +26,102 @@ only on subsequent ones (triggered by state or props changes).
...
+JavaScript:
+```js
+const { Component } = owl;
+const { Portal } = owl.misc;
+
+class TeleportedComponent extends Component {}
+class App extends Component {}
+
+App.components = { Portal , TeleportedComponent };
```
ged-odoo: In the doc, I usually use `static` keyword, and inline templates.

doc/reference/misc.md
```diff
@@ -21,3 +26,102 @@ only on subsequent ones (triggered by state or props changes).
...
+</body>
+```
+
+### Expected Behaviors
+
+The teleported piece is updated as any other `Component`'s DOM and in the same sequence.
+Namely the teleported piece will be updated in function of its parents components, and patched as
+a normal child.
+
+The events triggered by a child component will be stopped to not bubble outside of the `target`.
```
ged-odoo: you should clarify that we are talking about events emitted by the `trigger` method, i.e. _business events_

src/component/component.ts
```diff
@@ -165,6 +168,7 @@ export class Component<T extends Env, Props extends {}> {
       });
       depth = 0;
     }
+    this._triggerHook = this.env[portalSymbol as any] || null;
```
ged-odoo: I think that this should not be registered on the component.

First, it pollutes the name space (also, we have the convention of using two `_` for owl internal stuff in `Component`). Second, it is some additional work at the creation of a component (which is much more common than emitting events). So, in my opinion, the check should be done in the `trigger` method, every time an event is emitted (it is fine to do it each time, it's a very very small amount of work.

src/component/component.ts
```diff
@@ -478,7 +475,25 @@ export class Component<T extends Env, Props extends {}> {
       }
     }
   }
-
+  /**
+   * Private trigger method, allows to choose the component which triggered
+   * the event in the first place
+   * Returns the result of dispatchEvent or null if nothing has been done
```
ged-odoo: why? would it not be simpler to just not return anything?

src/component/fiber.ts
```diff
@@ -222,6 +222,7 @@ export class Fiber {
     if (this.target) {
       this.target.appendChild(this.component.el!);
       inDOM = document.body.contains(this.target);
+      this.component.env.qweb.trigger('dom-appended');
```
ged-odoo: we should add a reference page in the `doc/architecture` to document all _internal_ events used by owl.

src/misc/index.ts
```diff
@@ -0,0 +1 @@
+export const portalSymbol = Symbol();
```
ged-odoo: not a big fan of this file.  I would define it in `Component.ts`, since you use it in that file anyway.

Also, it should be something like this:
```js
Symbol('portal');
```
for a better description of its use

src/misc/portal.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export class Portal extends Component<any, any> {
+  static template = xml`<portal><t t-slot="default"/></portal>`;
+  static props = {
+    target: {
+      type: String,
+    }
+  }
+
+  doTargetLookUp: Boolean = true;
```
ged-odoo: boolean, not Boolean. There is a difference :)

doc/reference/misc.md
aab-odoo: or must?

doc/reference/misc.md
```diff
@@ -21,3 +26,102 @@ only on subsequent ones (triggered by state or props changes).
...
+
+### Usage
+
+The content it will teleport is defined within the `<Portal>` node and
+internally uses the `default` [Slot](component.md#slots).
+
+This slot only contains **one** node, which in turn can have as many children as necessary.
+
+The element under which the content will be teleported is represented as a selector
+by the `target` props which only accepts a string as its value.
```
aab-odoo: as value

src/component/component.ts
aab-odoo: :+1:

src/misc/portal.ts
```diff
@@ -0,0 +1,148 @@
...
+ * Portal
+ *
+ * The Portal component allows to render a part of a component outside it's DOM.
+ * It is for example useful for dialogs: for css reasons, dialogs are in general
+ * placed in a specific spot of the DOM (e.g. directly in the body). With the
+ * Portal, a component can conditionally specify in its tempate that it contains
+ * a dialog, and where this dialog should be inserted in the DOM.
+ *
+ * The Portal component ensures that the communication between the content of
+ * the Portal and its parent properly works: events reaching the Portal are
```
aab-odoo: business events (or OwlEvents)

src/misc/portal.ts
aab-odoo: also maybe explain what it does, like you did for other attributes

src/misc/portal.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+  /**
+   * At each DOM change, ensures that the structure is correct
+   */
+  __checkVNodeStructure(vnode: VNode) {
+    const children = vnode.children!;
+    let countRealNodes = 0;
+    for (let child of children) {
+      if ((child as VNode).sel)
```
aab-odoo: missing { }

aab-odoo: call npm run prettier btw

src/misc/portal.ts
```diff
@@ -0,0 +1,148 @@
...
+   * And mark the content's Components as mounted
+   */
+  __deployPortal() {
+    this.__checkTargetPresence();
+    const portalElm = this.portal!.elm!;
+    this.target!.appendChild(portalElm);
+  }
+  /**
+   * Removes from the DOM
+   * the element we have teleported
```
aab-odoo: you can write up to 80 char per line ^^

src/misc/portal.ts
```diff
@@ -0,0 +1,148 @@
...
+   * we render
+   */
+  __checkTargetPresence() {
+    if (!this.target || !document.contains(this.target)) {
+      throw new Error(`Could not find any match for "${this.props.target}"`);
+    }
+  }
+  /**
+   * Actually launch the element towards the outside target
+   * And mark the content's Components as mounted
```
aab-odoo: how?

src/misc/portal.ts
```diff
@@ -0,0 +1,148 @@
...
+      const parent = displacedElm.parentNode;
+      if (parent) {
+        parent.removeChild(displacedElm);
+      }
+    }
+    super.__destroy(parent);
+  }
+  /**
+   * Ensures the dom is "ready" upstream
+   * to secure teleportation
```
aab-odoo: don't understand the docstring, nor its relation to the function

src/misc/portal.ts
```diff
@@ -0,0 +1,148 @@
...
+    this.__checkVNodeStructure(vnode);
+    const shouldDeploy = !this.portal && !this.doTargetLookUp;
+
+    const portalPatch = this.portal ? this.portal : document.createElement(vnode.children[0].sel);
+    this.portal = patch(portalPatch, vnode.children![0] as VNode);
+    vnode.children = [];
+
+    if (!this.doTargetLookUp && !shouldDeploy) {
+      // Only on pure patching, provided the
+      // this.target's parent has not been de-mounted
```
aab-odoo: unmounted

src/misc/portal.ts
```diff
@@ -0,0 +1,148 @@
...
+      this.__checkTargetPresence();
```
aab-odoo: i'd do this before patching

tools/playground/samples.js
```diff
@@ -1444,6 +1444,111 @@ const FORM_XML = `<templates>
 </templates>
 `;
 
+const PORTAL_COMPONENTS = `
+// This shows the expected use case of Portal
+// Which is to implement something similar
```
aab-odoo: which

kebeclibre:
@aab-odoo 
@ged-odoo 
Branch cleaned
>> Evaluate the playground
aab thinks it's too complicated. I think it is quite fine and representative of what we aim at doing.

>> The parent.env stuff is ugly, though I think it is not bad

ged-odoo:
I agree with @aab-odoo  on the complexity of the playground, but i do not want to block this any longer.

Issue #476, opened 2019-11-19
Context: components using a context that are destroyed before being mounted do not stop listening


aab-odoo:

aab-odoo:
```js
  test.only("destroyed component before being mounted is inactive", async () => {
    const testContext = new Context({ a: 123 });

    class Child extends Component<any, any> {
      static template = xml`<span><t t-esc="contextObj.a"/></span>`;
      contextObj = useContext(testContext);
      willStart() {
        return makeDeferred();
      }
    }
    class Parent extends Component<any, any> {
      static template = xml`<div><Child t-if="state.flag"/></div>`;
      static components = { Child };
      state = useState({ flag: true });
    }

    const parent = new Parent();
    const prom = parent.mount(fixture);
    await nextTick();
    expect(testContext.subscriptions.update.length).toBe(1);
    parent.state.flag = false;
    await prom;
    expect(fixture.innerHTML).toBe("<div></div>");
    // kind of whitebox...
    // we make sure we do not have any pending subscriptions to the 'update'
    // event
    expect(testContext.subscriptions.update.length).toBe(0); // is still 1
  });
```

aab-odoo:
ping @seb-odoo

PR #477, opened 2019-11-19
Components destroyed before being mounted

aab-odoo:
Related to issue #476

### DIFF
--- a/src/component/directive.ts
+++ b/src/component/directive.ts
@@ -443,7 +443,7 @@ QWeb.addDirective({
     );
     ctx.addLine(`const fiber = w${componentID}.__owl__.currentFiber;`);
     ctx.addLine(
-      `def${defID}.then(function () {if (fiber.isCompleted) {return;} const vnode = fiber.vnode; pvnode.sel = vnode.sel; ${createHook}});`
+      `def${defID}.then(function () { if (fiber.isCompleted) { w${componentID}.destroy(); return; } const vnode = fiber.vnode; pvnode.sel = vnode.sel; ${createHook}});`
     );
     if (registerCode) {
       ctx.addLine(registerCode);
--- a/src/context.ts
+++ b/src/context.ts
@@ -2,7 +2,7 @@ import { Component } from "./component/component";
 import { scheduler } from "./component/scheduler";
 import { EventBus } from "./core/event_bus";
 import { Observer } from "./core/observer";
-import { onWillUnmount } from "./hooks";
+
 /**
  * The `Context` object provides a way to share data between an arbitrary number
  * of component. Usually, data is passed from a parent to its children component,
@@ -138,9 +138,11 @@ export function useContextWithCB(ctx: Context, component: Component<any, any>, m
       await method();
     }
   });
-  onWillUnmount(() => {
+  const __destroy = component.__destroy;
+  component.__destroy = (parent) => {
     ctx.off("update", component);
     delete mapping[id];
-  });
+    __destroy.call(component, parent);
+  }
   return ctx.state;
 }
--- a/tests/__snapshots__/animations.test.ts.snap
+++ b/tests/__snapshots__/animations.test.ts.snap
@@ -35,7 +35,7 @@ exports[`animations t-transition combined with component 1`] = `
         };
         utils.transitionRemove(vn, 'chimay', finalize);}}});
         const fiber = w3.__owl__.currentFiber;
-        def2.then(function () {if (fiber.isCompleted) {return;} const vnode = fiber.vnode; pvnode.sel = vnode.sel; });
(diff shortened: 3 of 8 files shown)


seb-odoo:
My stack when the "problematic" `__destroy` is called (but it is actually called, and if I do the fix in `context.ts` it solves the issue).

```
__destroy (owl.js:3698)
component.__destroy (use_dev_log.js:43)
__destroy (owl.js:3696)
component.__destroy (use_dev_log.js:43)
destroy (owl.js:3640)
component.destroy (use_dev_log.js:37)
(anonymous) (VM85410:60)
render (owl.js:1485)
__owl__.renderFn (owl.js:4165)
__render (owl.js:3844)
(anonymous) (owl.js:3616)
Promise.then (async)
render (owl.js:3612)
__owl__.observer.notifyCB (owl.js:4159)
set (owl.js:122)
_onDropZoneFilesDropped (attachment_box.js:72)
await in _onDropZoneFilesDropped (async)
eval (VM85410:46)
trigger (owl.js:3667)
_onDrop (drop_zone.js:117)
extra.handlers.<computed> (VM85412:17)
invokeHandler (owl.js:571)
handleEvent (owl.js:599)
handler (owl.js:604)
dropFiles (test_utils.js:237)
(anonymous) (attachment_box_tests.js:180)
await in (anonymous) (async)
runTest (qunit-2.9.1.js:3045)
```

The code in VM85410 calling destroy is:
```
        let w16 = templateId16 in parent.__owl__.cmap ? parent.__owl__.children[parent.__owl__.cmap[templateId16]] : false;
        let props16 = {areAttachmentsEditable:true,attachmentLocalIds:context['storeProps'].attachmentLocalIds};
        if (w16 && w16.__owl__.currentFiber && !w16.__owl__.vnode) {
            w16.destroy();
            w16 = false;
        }
```

seb-odoo:
The reason why it is destroyed in my case is because a store update happened right after it started to "render" (not sure if it is the proper term). In my case it is somewhat "wrongly" destroyed at this step, because the following "render" will also need to recreate it (in other words, it is ultimately a props change, but apparently this does not count as props change if no render was ever finished before).

seb-odoo:
I think the `__destroy` is not called in your case comparatively to mine because in my case, the parent is mounted once before.

It goes like this:
- parent is mounted without the child (because `t-if` is `false` by default)
- store state change makes the parent `render` again and so the child is constructed (because `t-if` becomes `true`)
- then the child is destroyed before it was mounted. This is due to a change of local state that happens asynchronously (it was waiting on promise which is likely resolved on the same tick as the store change or just after)
- note that the child is recreated again just after because `t-if` is still `true`

seb-odoo:
I managed to get a destroy with this code:

Though the final assert is failing even with the "current fix".

```js
test.only("destroyed component before being mounted is inactive", async () => {
    const testContext = new Context({ a: 123 });

    class Child extends Component<any, any> {
      static template = xml`<span><t t-esc="contextObj.a"/></span>`;
      contextObj = useContext(testContext);
      willStart() {
        const prom = makeDeferred();
        setTimeout(() => {
          parent.state.flag = true;
          prom.resolve();
        });
        return prom;
      }
      destroy() {
        console.warn('DESTROY');
      }
    }
    class Parent extends Component<any, any> {
      static template = xml`<div t-debug="1"><Child t-if="contextObj.a === 456"/></div>`;
      static components = { Child };
      contextObj = useContext(testContext);
      state = useState({ flag: false });
    }

    const parent = new Parent();
    const prom = parent.mount(fixture);
    await prom;
    expect(testContext.subscriptions.update.length).toBe(1);
    expect(fixture.innerHTML).toBe("<div></div>");

    testContext.state.a = 456;
    await nextTick();
    await nextTick();
    expect(fixture.innerHTML).toBe("<div><span>456</span></div>");
    // kind of whitebox...
    // we make sure we do not have any pending subscriptions to the 'update'
    // event
    expect(testContext.subscriptions.update.length).toBe(2);
  });
```

src/component/directive.ts
```diff
@@ -443,7 +443,7 @@ QWeb.addDirective({
     );
     ctx.addLine(`const fiber = w${componentID}.__owl__.currentFiber;`);
     ctx.addLine(
-      `def${defID}.then(function () {if (fiber.isCompleted) {return;} const vnode = fiber.vnode; pvnode.sel = vnode.sel; ${createHook}});`
+      `def${defID}.then(function () { if (fiber.isCompleted) { if (!w${componentID}.__owl__.isDestroyed) { w${componentID}.destroy(); } return; } const vnode = fiber.vnode; pvnode.sel = vnode.sel; ${createHook}});`
```
ged-odoo: you can simply call the `destroy` method, which will co the check anyway. we might as well try to keep the directive small

PR #481, opened 2019-11-20
[FIX] component: destroy not yet mounted components (2)

aab-odoo:
Following 2922cee6ea9b9f

Previous commit was not correct: used children with a cancelled
fiber (for instance, with a new currentFiber) could be destroyed.
This commit fixes the issue differently, by directly detecting
children that are not used anymore (and not mounted), and destroy
them directly (no need to wait for willStart promise to be resolved
anymore).


seb-odoo:
Good, I think I had that issue, I was having trouble debugging something random and since I updated to this revision I don't have it anymore.

Issue #473, opened 2019-11-18
Component having t-slot and changing state after mounted in t-foreach has incorrect value


Arcasias:
**JavaScript :**
```js
const { Component, tags, useState } = owl;

let activeDialog = null;

class Dialog extends Component {
    static template = tags.xml`
        <span t-att-class="{active: state.active}">
            <t t-slot="default"/>
        </span>`;
    state = useState({ active: false });
    mounted() {
        if (activeDialog) {
            activeDialog.state.active = false;
        }
        activeDialog = this;
        this.state.active = true;
    }
}

class App extends Component {
    static components = { Dialog };
    static template = tags.xml`
        <div>
            <button t-on-click="increment">Increment</button>
            <Dialog t-foreach="Array(state.count)" t-as="n" t-key="n_index" t-esc="n_index"/>
        </div>`;
    state = useState({ count: 0 });
    increment() {
        this.state.count ++;
    }
}

const app = new App();
app.mount(document.body);

// Add 3 items
(async () => {
    for (let i = 0; i < 3; i ++) {
        app.increment();
        await new Promise(r => Component.scheduler.requestAnimationFrame(() => setTimeout(r)));
    }
})();
```
**XML :**
```xml
<templates/>
```
**CSS (for readability) :**
```css
.active { color: red; }
span { padding: 5px; }
```
**Problem :**
Penultimate dialog consistently takes the value of the last one.

aab-odoo:
Issue with t-slot: scope and vars aren't correctly passed at subsequent (local) re-rendering
```js
const { Component, tags, useState } = owl;

class Dialog extends Component {
    static template = tags.xml`
        <span>
            <t t-esc="state.val"/>
            <t t-slot="default"/>
        </span>`;
    state = useState({ val: 'A' });
    mounted() {
        this.state.val = 'B';
    }
}

class App extends Component {
    static components = { Dialog };
    static template = tags.xml`
        <div>
            <Dialog t-foreach="Array(2)" t-as="n" t-key="n_index" t-esc="n_index"/>
        </div>`;
}

const app = new App();
app.mount(document.body);
```

produces: **B1 B1** instead of **B0 B1**

aab-odoo:
Issue with t-foreach:
```js
const { Component, tags, useState } = owl;

let activeDialog = null;

class Dialog extends Component {
    static template = tags.xml`
        <span t-att-class="{active: state.active}">
            dialog
        </span>`;
    state = useState({ active: false });
    mounted() {
[...]

PR #484, opened 2019-11-21
[FIX] component: reset fiber to null before patching

ged-odoo:
Before this commit, some components were reset, some other were not,
while patching the UI.  With this commit, we do reset the currentFiber
to null before starting the patching process.

This solves some weird issues such as the one shown in the test.

part of #473

### DIFF
--- a/src/component/component.ts
+++ b/src/component/component.ts
@@ -622,7 +622,7 @@ export class Component<T extends Env, Props extends {}> {
    * Only called by qweb t-component directive
    */
   __mount(fiber: Fiber, elm: HTMLElement): VNode {
-    if (fiber !== this.__owl__.currentFiber) {
+    if (fiber !== this.__owl__.currentFiber && this.__owl__.currentFiber) {
       fiber = this.__owl__.currentFiber!; // TODO: check if we can remove fiber arg
     }
     const vnode = fiber.vnode!;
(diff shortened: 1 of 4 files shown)


src/component/component.ts
```diff
@@ -622,7 +622,7 @@ export class Component<T extends Env, Props extends {}> {
    * Only called by qweb t-component directive
    */
   __mount(fiber: Fiber, elm: HTMLElement): VNode {
-    if (fiber !== this.__owl__.currentFiber) {
+    if (fiber !== this.__owl__.currentFiber && this.__owl__.currentFiber) {
       fiber = this.__owl__.currentFiber!; // TODO: check if we can remove fiber arg
     }
     const vnode = fiber.vnode!;
```
aab-odoo: move line 629 above

tests/component/component.test.ts
```diff
@@ -2423,6 +2422,43 @@ describe("random stuff/miscellaneous", () => {
...
+    }
+
+    const widget = new Parent();
+    await widget.mount(fixture);
+    expect(fixture.innerHTML).toBe("<div><span>0</span></div>");
+    await nextTick();
+    expect(fixture.innerHTML).toBe("<div><span>1</span></div>");
+    widget.state.push(true);
+    await nextTick();
+    expect(fixture.innerHTML).toBe("<div><span>0</span><span>1</span></div>");
```
aab-odoo: maybe check the number of calls to __render, just for fun?

ged-odoo:
so, to be clear, this is not correct. another PR is working on this...

PR #492, opened 2019-11-21
Master fiber change 2

ged-odoo:


ged-odoo:
I just rebased on master and force pushed

PR #511, opened 2019-11-27
fix usestore update & destroy

seb-odoo:


seb-odoo:
Ready for me, requesting review on these changes

src/store.ts
```diff
@@ -95,37 +95,42 @@ export function useStore(selector, options: SelectorOptions = {}): any {
   if (!store.updateFunctions[component.__owl__.id]) {
```
ged-odoo: maybe you could assign componentId above this line, it is used 3 times now...

seb-odoo: I was thinking that too but I already pushed the commit... I'll change it then.

seb-odoo: Rebased, added componentId, and replaced all the occurrence in the function

PR #517, opened 2019-11-28
[IMP] component: allow subscribing to scheduler updates

seb-odoo:
Please consider the following PR, to properly subscribe to scheduler updates.

### DIFF
--- a/doc/learning/testing_components.md
+++ b/doc/learning/testing_components.md
@@ -108,28 +108,11 @@ make sure that the DOM is up-to-date.
 
 It is sometimes useful to wait until Owl is completely done updating components
 (in particular, if we have a highly concurrent user interface). This next
-helper simply polls every 20ms the internal Owl task queue and returns a promise
-which resolves when it is empty:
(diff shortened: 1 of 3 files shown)


seb-odoo:
Ready for review, but I'm not particularly happy about the part where I copied makeDeferred into scheduler, if you have a better suggestion.

gdeb:
Could you tell me what problem you are trying to solve? I am not fond of changing that much the code for the sake of tests

seb-odoo:
It is for test, with the current suggested helper, I could not find a way to make sure the queue was "busy" at least once between when I called the helper and when it resolved, which would ensure me that something was done and that it wasn't just a dummy call.

I came close, but I had cases where the queue was initially empty, then a task was done, and then queue was empty again before I could notice it. The promise would resolve, but I would consider it as a dummy call when in fact it wasn't.

I don't have the issue anymore with the current PR.

ged-odoo:
sorry, but I think that a better solution is to to it externally like we talked.

PR #518, opened 2019-11-28
[FIX] qweb: cascading t-call and t-raw="0"

### LINKED ISSUE #510: Cascading t-call and t-raw="0"
```
  <div t-name="App" >
    <t t-call="Dialog">
       <div>RAW</div> 
    </t>
  </div>

  <div t-name="Dialog">
    <t t-call="Modal" >
      <t t-raw="0" />
    </t>
  </div>

  <div t-name="Modal">
      <t t-raw="0" />
  </div>
```

Crashes with 
```
Maximum call stack size exceeded
    at Function.from (<anonymous>)
    at QWeb._compileChildren (owl.js:1921)
    at QWeb._compileNode (owl.js:1751)
```

kebeclibre:
Have a t-call having a t-raw="0"
that call is made to a template of the same form
i.e. itself as a t-call t-raw="0" structure

Before this commit, there was recursion crash
This was because the caller to which a t-raw="0" referred to
was incorrect. The caller here is the node which makes the t-call.
In particular, the caller was always set to last caller of the context

After this commit, there is no crash and this imbrication
of t-call and t-raw="0" is well rendered.
To sum up, we count the recursive calls to t-raw="0" and fetch
the caller in the context accordingly

closes #510

### DIFF
--- a/src/qweb/compilation_context.ts
+++ b/src/qweb/compilation_context.ts
@@ -32,6 +32,8 @@ export class CompilationContext {
   scopeVars: any[] = [];
   currentKey: string = "";
   templates: { [key: string]: boolean } = {};
+  callingLevel: number = 0;
+  inliningLevel: number = 0;
 
   constructor(name?: string) {
     this.rootContext = this;
@@ -129,6 +131,10 @@ export class CompilationContext {
   subContext(key: keyof CompilationContext, value: any): CompilationContext {
     const newContext = Object.create(this);
     newContext[key] = value;
+    if (key === 'caller') {
+      newContext.callingLevel++;
+      newContext.inliningLevel++;
+    }
     return newContext;
   }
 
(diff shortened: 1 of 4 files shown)


src/qweb/compilation_context.ts
```diff
@@ -166,6 +172,19 @@ export class CompilationContext {
...
   }
+  /**
+   * Recursively (inverse) fetches the `caller` of a context
+   * Useful to determine to which t-call a t-raw="0" refers
+   */
+  getCaller(targetLevel: number) {
+    if (targetLevel === this.callingLevel) {
+      return this.caller;
+    }
+    if ((this as any).__proto__) {
```
ged-odoo: nitpick:
```js
const proto = (this as any).__proto__;
return proto ? proto.getCaller(targetLevel): null;
```

src/qweb/compilation_context.ts
```diff
@@ -166,6 +172,19 @@ export class CompilationContext {
     this.dedent();
     this.addLine("}");
   }
+  /**
+   * Recursively (inverse) fetches the `caller` of a context
+   * Useful to determine to which t-call a t-raw="0" refers
+   */
+  getCaller(targetLevel: number) {
```
ged-odoo: add return value in signature
```js
  getCaller(targetLevel: number): CompilationContext | null {
```

ged-odoo:
looking good. i just made one or two comments, then we squash and merge this...

PR #527, opened 2019-12-02
[FIX] component: tricky concurrency issue

### LINKED ISSUE #520: event handler of parent not called when triggering event from children in some situations
It seems (without guarantee) to be related to nested components, where the child is triggering the event, and the parent was initially mounted, then destroyed, and recreated again.

ged-odoo:
closes #520


aab-odoo:
nice

PR #534, opened 2019-12-03
[FIX] qweb: properly handle empty class attribute

### LINKED ISSUE #530: empty class causes a crash
```js
    <div class=" ">test</div>
```

same with "" or "   "

ged-odoo:
The classList.add method actually crashes when given an empty string (or
a string with just white spaces).

closes #530

### DIFF
--- a/src/qweb/qweb.ts
+++ b/src/qweb/qweb.ts
@@ -650,13 +650,14 @@ export class QWeb extends EventBus {
       if (!name.startsWith("t-") && !(<Element>node).getAttribute("t-attf-" + name)) {
         const attID = ctx.generateID();
         if (name === "class") {
-          let classDef = value
-            .trim()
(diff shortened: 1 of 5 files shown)


src/qweb/qweb.ts
```diff
@@ -650,13 +650,15 @@ export class QWeb extends EventBus {
...
         const attID = ctx.generateID();
         if (name === "class") {
-          let classDef = value
-            .trim()
-            .split(/\s+/)
-            .map(a => `'${a}':true`)
-            .join(",");
-          classObj = `_${ctx.generateID()}`;
-          ctx.addLine(`let ${classObj} = {${classDef}};`);
+          if (value.trim()) {
```
aab-odoo: maybe do `value = value.trim();` before if() to prevent from calling trim twice

PR #536, opened 2019-12-03
[FIX] portal: manually unmounting works

### LINKED ISSUE #531: Portal is not destroyed when parent is unmounted
**JavaScript :**
```js
const { Portal } = owl.misc;

    static components = { Portal };
}

const app = new App();
app.mount(document.body)
    .then(() => app.unmount());
```
**XML :**
```xml
<templates>
  <div t-name="App">
    <Portal target="'body'">
      <p>I'm travelling through space!</p>
    </Portal>
  </div>
</templates>
```
**Error :**
`Portal` is not destroyed after `App` is unmounted

kebeclibre:
Before this commit, when manually unmounting the portal
or one of its parents, the teleported elements were not cleaned
and stayed teleported forever

After this commit, the teleported content is brought back and
unmounted much like any component

closes #531

### DIFF
--- a/src/misc/portal.ts
+++ b/src/misc/portal.ts
@@ -60,6 +60,16 @@ export class Portal extends Component<any, any> {
     });
   }
   /**
+   * Override to revert back to a classic Component's structure
+   *
+   * @override
+   */
+  __callWillUnmount() {
+    super.__callWillUnmount();
+    this.el!.appendChild(this.portal!.elm!);
+    this.doTargetLookUp = true;
+  }
+  /**
    * At each DOM change, we must ensure that the portal contains exactly one
    * child
    */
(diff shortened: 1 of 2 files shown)


src/misc/portal.ts
```diff
@@ -59,6 +59,21 @@ export class Portal extends Component<any, any> {
...
+   */
+  __callWillUnmount() {
+    if (this.el && this.portal && this.portal.elm) {
+      this.el.appendChild(this.portal.elm);
+      const vnode = this.__owl__.vnode;
+      vnode!.children = [this.portal];
+      this.portal = null;
+      this.doTargetLookUp = true;
+    }
+    super.__callWillUnmount();
```
ged-odoo: this should be done at the beginning. otherwise subcomponents will not be at the correct place when they are unmounted

kebeclibre:
@ged-odoo 
Done !
Hooks willunmount and mounted are correctly tested

src/misc/portal.ts
```diff
@@ -59,6 +59,18 @@ export class Portal extends Component<any, any> {
...
     });
   }
+  /**
+   * Override to revert back to a classic Component's structure
+   *
+   * @override
+   */
+  __callWillUnmount() {
+    super.__callWillUnmount();
+    if (this.el && this.portal && this.portal.elm) {
```
ged-odoo: I think that if you component is being unmounted, then it necessarily has a `el`, `portal` and `portal.elm`, so you can remove this `if`

PR #557, opened 2019-12-06
[FIX] slot: nested t-slot multiple levels

### LINKED ISSUE #545: nested t-slot memory error
```js
    class Child3 extends Widget {
      static template = xml`
        <span>Child 3</span>`;
    }
    class Modal extends Widget {
      static template = xml`
        <span id="modal">
          <t t-slot="default"/>
        </span>`;
    }
    class Portal extends Widget {
      static template = xml`
        <span id="portal">
          <t t-slot="default"/>
        </span>`;
    }
    class Dialog extends Widget {
      static components = { Modal, Portal };
      static template = xml`
        <span id="c1">
          <Modal>
            <Portal>
              <t t-slot="default"/>
            </Portal>
           </Modal>
        </span>`;
    }
    class Parent extends Widget {
      static components = { Child3, Dialog }
      static template = xml`
        <span id="c1">
          <Dialog>
            <Child3/>
          </Dialog>
        </span>`;
    }

```
This crashes in a browser with "Paused before out of memory crash"

It may be strongly related to #544

kebeclibre:
When a Component was nested in a slot within another directly
that is, in the same template, its rendering context was wrongly
its *parent*, whereas it should be the component which we *are* rendering
the template of.

A similar issue happened when a t-slot was nested within a slot, itself
having a t-slot.
In that case though, the crash in a browser was a memory error

This commit fixes both issues by introducing the concept of contextualParent
which is the Component of which we are compiling the template

At runtime (after compile time) this contextualParent is irrelevant
it doesn't take any part in the life cycle of any component

closes #545

### DIFF
--- a/src/component/directive.ts
+++ b/src/component/directive.ts
@@ -405,6 +405,9 @@ QWeb.addDirective({
       `if (!W${componentID}) {throw new Error('Cannot find the definition of component "' + componentKey${componentID} + '"')}`
     );
     ctx.addLine(`w${componentID} = new W${componentID}(parent, props${componentID});`);
+    if (ctx.currentComponent) {
+      ctx.addLine(`w${componentID}.__owl__.contextualParentID = ${ctx.currentComponent.__owl__.id};`);
+    }
     if (transition) {
       ctx.addLine(`const __patch${componentID} = w${componentID}.__patch;`);
       ctx.addLine(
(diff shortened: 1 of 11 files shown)


src/component/directive.ts
```diff
@@ -428,7 +425,10 @@ QWeb.addDirective({
           const key = slotNode.getAttribute("t-set")!;
           slotNode.removeAttribute("t-set");
           const slotFn = qweb._compile(`slot_${key}_template`, slotNode, ctx);
-          QWeb.slots[`${slotId}_${key}`] = slotFn;
+          QWeb.slots[`${slotId}_${key}`] = function ctxSlotFn(context, extra) {
+            context = ctx.currentComponent ? ctx.currentComponent : context;
```
ged-odoo: I am afraid this is wrong (but previous code was wrong as well). This means that the compiled function depends on the rendering context, which is the first component rendered.  I am pretty sure that this means that if the same template is rendered with a different component, you will get a wrong component

PR #558, opened 2019-12-08
Example work with owl.js

MohsenWaleed:
The file structure use owl.min.js and the xml code use the owl.js so it's better to change the file structure as suggested.

There is a comment in tutorial if anybody prefer to use owl.min.js so he will change the file name in xml.


Ewetoye (commented):
It is best left as is.

PR #579, opened 2019-12-13
[FIX] store: make Array.map work correctly on the result of selector

### LINKED ISSUE #577: "in" operator not working on the result of useStore
Fix in the returned Proxy of useStore:

```js
            has(target, k) {
                return k in result;
            },
```

This fixes it, but for some reason I can't reproduce the issue on a simple test.

In my code: the selector returns an array from the store (so the result of useStore is effectively a Proxy of Proxy).

seb-odoo:
closes #577


seb-odoo:
PR good to go for me.

1 concern: are other handlers from Proxy missing too?

PR #590, opened 2019-12-20
[IMP] component/tags: add inline css tag

### LINKED ISSUE #284: simple single file component for owl
With tagged templates, we can execute some code whenever a tagged literal is defined. Also, bonus point, it is possible to make it recognized by editor plugins such as vscode (see lit-html) for proper syntax highlighting and autocompletion.

So, it would be possible to have something like this, in a js/ts file:
```js
import { Component, css, xml } from 'owl/component';

// CSS
const styles = css`
  .my-component {
    display: block;
    background-color: 'red';
  }
`;

// XML
const template = xml`
  <div>
    Hello Owl
  </div>
`;

// JS
class MyComponent extends Component {
  static styles = styles;
  static template = template;
   
  ....
}
```

As a bonus, it means that we can inject variables from the JS into the CSS!

Then, many questions would need to be answered: do we collect all the css and simply add it to a stylesheet? Or do we only do it whenever a component is mounted? Do we prefix it with unique id to make it scoped to the component?  How do we do that?

ged-odoo:
closes #284

### DIFF
--- a/doc/reference/tags.md
+++ b/doc/reference/tags.md
@@ -4,16 +4,19 @@
 
 - [Overview](#overview)
 - [`xml` tag](#xml-tag)
+- [`css` tag](#css-tag)
 
 ## Overview
 
-Tags are very small helpers to make it easy to write inline templates. There is
-only one currently available tag: `xml`, but we plan to add other tags later,
-such as a `css` tag, which will be used to write [single file components](../tooling.md#single-file-component).
+Tags are very small helpers intended to make it easy to write inline templates
+or styles. There are currently two tags: `css` and `xml`. With these functions,
+it is possible to write [single file components](../tooling.md#single-file-component).
 
 ## XML tag
 
-Without tags, creating a standalone component would look like this:
+The `xml` tag is certainly the most useful tag. It is used to define an inline
+QWeb template for a component. Without tags, creating a standalone component
+would look like this:
 
 ```js
 import { Component } from 'owl'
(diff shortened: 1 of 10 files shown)


doc/reference/tags.md
```diff
@@ -52,3 +55,84 @@ class MyComponent extends Component {
...
+  static style = css`
+    .my-component {
+      color: ${theme.MAIN_COLOR};
+      background-color: ${theme.SECONDARY_color};
+    }
+  `;
+}
+```
+
+2. scoping rules to the current component:
```
aab-odoo: the same would be achieved with a classname or id right?

doc/tooling.md
```diff
@@ -43,32 +43,43 @@ useful to compare various performance metrics on some tasks.
 It is very useful to group code by feature instead of by type of file. It makes
 it easier to scale application to larger size.
 
-To do so, Owl currently has a small helper that makes it easy to define a
-template inside a javascript (or typescript) file: the [`xml`](reference/tags.md#xml-tag)
-helper. With this, a template is automatically registered to [QWeb](reference/qweb_engine.md).
+To do so, Owl has two small helpers that makes it easy to define a
```
aab-odoo: that make

tests/component/styles.test.ts
```diff
@@ -0,0 +1,132 @@
...
+import { makeTestFixture, makeTestEnv } from "../helpers";
+
+//------------------------------------------------------------------------------
+// Setup and helpers
+//------------------------------------------------------------------------------
+
+// We create before each test:
+// - fixture: a div, appended to the DOM, intended to be the target of dom
+//   manipulations.  Note that it is removed after each test.
+// - env: a WEnv, necessary to create new components
```
aab-odoo: env: an Env

doc/reference/tags.md
ged-odoo: yes, but not exactly.  Generating a unique id guarantees that you do not clash with the same classname somewhere else in your application.  It may not be a big issue if you write an application, but for a library, it is a more important issue.  Imagine a generic spreadsheet component written in an owl library.  If this component uses a classname such a `toolbar`, or `icon`,  or whatever, there is a big chance for a conflict.  So, we need a longer and maybe not so natural classname, such as `owl-spreadsheet-toolbar`.

But yeah, that is not really a big issue.

aab-odoo: mmmmh good point

Issue #592, opened 2019-12-24
Css tag nested rules


mcm-odoo:
With owl.tags.css, this
```css
div {
  [role=button], button {
    color: red;
  }
}
```
results in
```css
div [role=button], button {
  color: red;
}
```
it should be
```css
div [role=button], div button {
  color: red;
}
```

mcm-odoo:
I wanted to fix it so I tried to create a PR but I do not have the right to push my branch
Here the diff patch

```diff
@@ -9,12 +9,23 @@ export const STYLESHEETS: { [id: string]: HTMLStyleElement } = {};
 
 function processSheet(str: string): string {
   const tokens = str.split(/(\{|\}|;)/).map(s => s.trim());
-  const selectorStack: string[] = [];
+  const selectorStack: string[][] = [];
   const parts: string[] = [];
   let rules: string[] = [];
+  function generateSelector(stackIndex: number, parentSelector?: string) {
+    const parts: string[] = [];
+    for (const selector of selectorStack[stackIndex]) {
+      let part = parentSelector && parentSelector + ' ' + selector || selector;
+      if (stackIndex < selectorStack.length - 1) {
+        part = generateSelector(stackIndex + 1, part);
+      }
+      parts.push(part);
+    }
+    return parts.join(', ');
+  }
   function generateRules() {
     if (rules.length) {
-      parts.push(selectorStack.join(" ") + " {");
+      parts.push(generateSelector(0) + " {");
       parts.push(...rules);
       parts.push("}");
       rules = [];
@@ -28,7 +39,7 @@ function processSheet(str: string): string {
     } else {
       if (tokens[0] === "{") {
         generateRules();
-        selectorStack.push(token);
+        selectorStack.push(token.split(/\s*,\s*/));
         tokens.shift();
       }
       if (tokens[0] === ";") {

```

ged-odoo:
hey @mcm-odoo 

You actually don't need to push to the owl repo to create a PR. You can simply create a branch on your own repo (cloned from this one), and then you can submit a PR

PR #593, opened 2019-12-26
[FIX] styles: fix rule selector generation

mcm-odoo:
This PR fixes the css selector generation for nested rules

Before this fix, a selector like:

```css
.parent {
  .child-a, .child-b {
    color: red;
  }
}
```

was generated as:

```css
.parent .child-a, .child-b {
  color: red;
}
```

Now it generates:

```css
.parent .child-a, .parent .child-b {
  color: red;
}
```

#592

### DIFF
--- a/src/component/styles.ts
+++ b/src/component/styles.ts
@@ -7,14 +7,25 @@
 
 export const STYLESHEETS: { [id: string]: HTMLStyleElement } = {};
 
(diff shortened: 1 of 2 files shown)


mcm-odoo:
thanks @ged-odoo 

I have added the test, it works correctly.

PR #597, opened 2019-12-30
[FIX] qweb, extensions: t-foreach needs to hold on scope

### LINKED ISSUE #594: T-foreach loop values are unavailable in inline handlers
**JavaScript :**
```js
    state = owl.useState({
        values: ['a'],
    });
}

const app = new App();
app.mount(document.body);
```

**XML :**
```xml
<templates>
  <div t-name="App">
    <div t-foreach="state.values" t-as="val" t-key="val">
      <!-- Available here -->
      <t t-esc="val_index"/>: <t t-esc="val + ''"/>
      <!-- ...but not here -->
      <button t-on-click="state.values.push(val)">Add same value</button>
    </div>
  </div>
</templates>
```
**Issue :**
None of the variables created during the compilation of the `t-foreach` directive (`$as`, `$as_index`, etc.) are accessible in inline event handlers.

kebeclibre:
Have a t-on within a t-foreach
the t-on has an expression

Before this commit, when triggering the t-on, the expression was falsely evaluated
In details, the expression took scope from outside the foreach loop
as we protect it to have similar results than an actual for loop

But, at triggering time, the scope protection had already been terminated
meaning the info of the iteration of the loop the handler has been built in
had already disappeared

This commit fixes that

closes #594

### DIFF
--- a/tests/component/component.test.ts
+++ b/tests/component/component.test.ts
@@ -2670,6 +2670,60 @@ describe("other directives with t-component", () => {
     await nextTick(); // wait for changes triggered in mounted to be applied
     expect(fixture.innerHTML).toBe("<div><span>B0</span><span>B1</span></div>");
   });
+
+  test("t-on expression in t-foreach", async () => {
+    class SomeWidget extends Component<any, any> {
+      static template = xml`
+      <div>
+        <div t-foreach="state.values" t-as="val" t-key="val">
+          <t t-esc="val_index"/>: <t t-esc="val + ''"/>
+          <button t-on-click="otherState.vals.push(val)">Expr</button>
+        </div>
+      </div>`;
+
+      state = useState({values: ['a', 'b']});
+      otherState = {vals: []};
+    }
+    const widget = new SomeWidget();
+    await widget.mount(fixture);
+
+    expect(fixture.innerHTML).toBe("<div><div>0: a<button>Expr</button></div><div>1: b<button>Expr</button></div></div>");
+    expect(widget.otherState.vals).toStrictEqual([]);
+    const buttons = fixture.querySelectorAll('button');
+    buttons[0].click();
+    buttons[1].click();
+    expect(widget.otherState.vals).toStrictEqual(['a', 'b']);
+    expect(QWeb.TEMPLATES[SomeWidget.template].fn.toString()).toMatchSnapshot();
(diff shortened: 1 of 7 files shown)


kebeclibre:
@ged-odoo 
Note that making a function call is not affected by the scope being reset because there is a block-specific var created here: https://github.com/odoo/owl/blob/master/src/qweb/extensions.ts#L62

I test this use case anyway, but is not affected by my fix

We may adopt this same approach (as in https://github.com/odoo/owl/blob/master/src/qweb/extensions.ts#L62)  instead of the fix I propose, but I think that my fix may prove useful in other cases of block-scoping e.g. t-if etc.......
Those would need more testing I guess :D

tests/component/component.test.ts
```diff
@@ -2670,6 +2670,56 @@ describe("other directives with t-component", () => {
     await nextTick(); // wait for changes triggered in mounted to be applied
     expect(fixture.innerHTML).toBe("<div><span>B0</span><span>B1</span></div>");
   });
+
+  test("t-on expression in t-foreach", async () => {
+    env.qweb.addTemplate("SomeWidget",
```
ged-odoo: this is a very old style of writing tests. I prefer using the `xml` tag helper instead. You most likely copied some code from an old test.

tests/component/component.test.ts
```diff
@@ -2670,6 +2670,56 @@ describe("other directives with t-component", () => {
...
+
+    expect(fixture.innerHTML).toBe("<div><div>0: a<button>Expr</button></div></div>");
+    expect(widget.otherState.vals).toStrictEqual([]);
+    fixture.querySelector('button')!.click();
+    expect(widget.otherState.vals).toStrictEqual(['a']);
+    expect(env.qweb.templates.SomeWidget.fn.toString()).toMatchSnapshot();
+  });
+
+  test("t-on method call in t-foreach", async () => {
+    env.qweb.addTemplate("SomeWidget",
```
ged-odoo: same here, I prefer inline templates. It makes it easier to read/understand.

kebeclibre:
https://github.com/odoo/owl/pull/599/commits/0e2e75084b18fd905cae5dbdcec08ff835008341

ged-odoo:
-

PR #601, opened 2020-01-07
[FIX] qweb: code block scoping  (foreach)

kebeclibre:
This PR puts together the main principles of #599 and #597.
The two issues are treated differently, in two separate commits.
But their reunion in one PR allows for consistency

### DIFF
--- a/src/qweb/compilation_context.ts
+++ b/src/qweb/compilation_context.ts
@@ -17,6 +17,7 @@ export class CompilationContext {
   rootContext: CompilationContext;
   shouldDefineParent: boolean = false;
   shouldDefineScope: boolean = false;
+  protectedScopeNumber: number = 0;
   shouldDefineQWeb: boolean = false;
   shouldDefineUtils: boolean = false;
   shouldDefineRefs: boolean = false;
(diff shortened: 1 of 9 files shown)


src/qweb/compilation_context.ts
```diff
@@ -156,6 +156,20 @@ export class CompilationContext {
...
+  captureExpression(expr: string): string {
+    const argId = this.generateID();
+    this.rootContext.shouldDefineScope = true;
+    const tokens = compileExprToArray(expr, this.variables);
+    const done = new Set();
+    return tokens.map((tok) => {
+      if (tok.varName && !done.has(tok.varName)) {
+        done.add(tok.varName);
+        this.addLine(`const ${tok.varName}_${argId} = ${tok.value};`);
+        tok.value = `${tok.varName}_${argId}`;
```
ged-odoo: this is not correct: the `const` should only be done once, but the `tok.value` replacement should be done for every tokens in the expression

src/qweb/extensions.ts
```diff
@@ -67,8 +67,10 @@ export function makeHandlerCode(
     }
   } else {
     // if we get here, then it is an expression
+    // we need to capture every variable in it
     putInCache = false;
-    code = ctx.formatExpression(value);
+    ctx.rootContext.shouldDefineScope = true;
```
ged-odoo: no need to define scope here, it is done in `captureExpression`

src/qweb/qweb.ts
```diff
@@ -122,10 +122,17 @@ const UTILS: Utils = {
       .join("");
   },
   getComponent(obj) {
-    while (obj && !obj.hasOwnProperty("__owl__")) {
+    return this.getScope(obj, "__owl__");
+  },
+  getScope(obj, str: string) {
```
ged-odoo: this is only used to write to the scope. if you look at its uses, there is always some duplication.  I think that this should probably be changed to a function `writeToScope(obj, str: string, value: value)`, so it is much clearer what it does, and can be slightly optimized.

ged-odoo:
this is really good work

src/qweb/compilation_context.ts
kebeclibre: done + tested

src/qweb/extensions.ts
kebeclibre: done

src/qweb/qweb.ts
kebeclibre: WritetoScope as suggested cannot work with ++var, because var gets *assigned* to the current scope before resolving in which scope it should be written in.

kebeclibre:
@ged-odoo 
Done, up for second round of review

Issue #591, opened 2019-12-21
css hook: support & and self


ged-odoo:
It would be very useful to be able to do this:

```js
static style = css`
    self {
       text-color: red;
       .some-class {
         font-size: 15px;
         &.vertical {
            font-weight: bold;
         }
       }
    }`; 
```
- `&`: useful to join a subrule with a parent selector,
- `self`: to select the root of the component (in which case, it will add a unique class to the component and use it instead of `self` in the actual sheet)

mcm-odoo:
For `self` we could add a new static property in the component `styleTarget` or `selector` (or an other name) which will define "the unique class". Like this the dev could set a custom class or maybe an attribute?

e.g.
```js
class ColoredComp extends Component {
  static style = css`self { color: red }`;
  static selector = '.my-colored-component'; // or an attribute [component=colored]
  static template = xml`<div>I'm red</div>`;
}
```
will result in dom
```html
<div class="my-colored-component">I'm red</div>
<!-- Or with attribute -->
<div component="colored">I'm red</div>
```

And also with this property we could do something like
```js
class FieldChar extends Component {
  static selector = '.field-char';
}

class FieldNumber extends Component {
  static selector = '[field=number]';
}

class List extends Component {
  static style = css`
    self {
      .field-char {
        add a special style when component FieldChar is in List
      }
      [field=number] {
        add a special style when component FieldNumber is in List
      }
    }`;
}

class Form extends Component {
  static style = css`
    self {
      .field-char {
        add a special style when component FieldChar is in Form
      }
      [field=number] {
        add a special style when component FieldNumber is in Form
      }
    }`;
}
```
This would give a special style to FieldChar and FieldNumber depending they are in Form or List.

ged-odoo:
fixed in owl 2, by removing the css tag

PR #606, opened 2020-01-08
Handle & selector

mcm-odoo:
This PR adds the support of the & selector.
This selector is useful to join a subrule with a parent selector.

example:
```scss
button {
  &:hover {
    background-color: red;
  }
}
```
will give
```css
button:hover {
  background-color: red;
}
```

part of https://github.com/odoo/owl/issues/591


ged-odoo:
missing doc

PR #607, opened 2020-01-08
[FIX] debug tools: do not crash on non json-stringifiable props

### LINKED ISSUE #595: debug script: cannot handle some props
if a props is non-stringifiable, it crashes.

ged-odoo:
closes #595


tools/debug.js
```diff
@@ -35,7 +35,12 @@
   });
 
   function toStr(obj) {
-    let str = JSON.stringify(obj || {});
+    let str;
+    try {
+      str = JSON.stringify(obj || {});
+    } catch (e) {
+      str = "<...>";
```
kebeclibre: @ged-odoo 
Might be ambiguous with `...` in case the string is too long
I'd go for for something explicit e.g. `<JSON error>`

aab-odoo: I agree

PR #621, opened 2020-01-18
[ADD] tools: add single file component example to playground

ged-odoo:


tools/playground/samples.js
```diff
@@ -1811,6 +1811,48 @@ const WMS_CSS = `body {
     font-size: 20px;
 }`;
 
+const SFC = `// This example illustrates how Owl enables single file components,
+// which includes code, template and style.
```
aab-odoo: include

tools/playground/samples.js
```diff
@@ -1811,6 +1811,48 @@ const WMS_CSS = `body {
...
+//
+// This is very useful in some situations, such as testing or quick prototyping.
+// Note that this example has no xml or css.
```
aab-odoo: no xml or css?

PR #630, opened 2020-02-05
[IMP] types: make component generic types optional

ged-odoo:
In most cases, we just want Component<any, Env>. But since it was so
annoying to have always the type Env, we actually used
Component<any,any> everywhere.

With this commit, the generic types have a default (and their order is
swapped), so we can simply use Component in most cases, and
Component<Props> when we want to type the props.


ged-odoo:
@aab-odoo I think this is a very large improvement in ease of use for typescript users

PR #644, opened 2020-02-16
[FIX] component: cancel previous mounting operations if necessary

### LINKED ISSUE #626: mounting on two different targets
If I mount manually a component on two different target, the first mount return promise should fail

ged-odoo:
closes #626

### DIFF
--- a/src/component/component.ts
+++ b/src/component/component.ts
@@ -37,8 +37,10 @@ export interface Env {
   [key: string]: any;
 }
 
+export type MountPosition = "first-child" | "last-child" | "self";
+
 interface MountOptions {
-  position?: "first-child" | "last-child" | "self";
+  position?: MountPosition;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


tests/component/un_mounting.test.ts
```diff
@@ -406,4 +396,46 @@ describe("unmounting and remounting", () => {
     await w.mount(span);
     expect(fixture.innerHTML).toBe("<div></div><span><div>Hey</div></span>");
   });
+
+  test("widget can be mounted on different target, another situation", async () => {
```
aab-odoo: targets

src/component/component.ts
```diff
@@ -311,6 +311,9 @@ export class Component<Props extends {} = any, T extends Env = Env> {
         return Promise.resolve();
       }
     }
+    if (__owl__.currentFiber) {
+      scheduler.rejectFiber(__owl__.currentFiber, "Mounting operation cancelled");
```
aab-odoo: no check on the target?

PR #667, opened 2020-03-02
[FIX] component: self mounting position keeps the reference

kebeclibre:
Have

```xml
<body t-name="webclient" />
```
and
```js

const comp = new WebClient();
comp.mount(document.body, {position: 'self'});
```

Before this commit, the body that was there before anything had happened
was *replaced* by the new body node created by the WebClient OWL component

After this commit, we ensure that the element body is the same at reference level

### DIFF
--- a/src/component/fiber.ts
+++ b/src/component/fiber.ts
@@ -225,6 +225,12 @@ export class Fiber {
               `Cannot attach '${component.constructor.name}' to target node (not same tag name)`
             );
           }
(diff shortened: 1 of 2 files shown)


kebeclibre:
patch is naive, we need to talk :D

kebeclibre:
@aab-odoo 
@ged-odoo 
The patch is now mature I think

It relies entirely on the heuristics of VNode.sameVnode though, I don't know to which point the patch is robust.....

Issue #672, opened 2020-03-11
t-set a recursive index within t-foreach and t-call


kebeclibre:
Given the test:

```js
  test("recursive template, part 4: with t-set recursive index", () => {
    qweb.addTemplates(`
        <templates>
            <div t-name="Parent">
                <t t-call="nodeTemplate">
                    <t t-set="recursive_idx" t-value="1"/>
                    <t t-set="node" t-value="root"/>
                </t>
            </div>
            <div t-name="nodeTemplate">
                <t t-set="recursive_idx" t-value="recursive_idx + 1"/>
                <p><t t-esc="node.val"/> <t t-esc="recursive_idx"/></p>
                <t t-foreach="node.children or []" t-as="subtree">
                    <t t-call="nodeTemplate">
                        <t t-set="node" t-value="subtree"/>
                    </t>
                </t>
            </div>

        </templates>
    `);
    const root = { val: "a", children: [{ val: "b", children: [{ val: "c" , children: [{val: "d"}] }] }] };
    const expected =
      "<div><div><p>a 2</p><div><p>b 3</p><div><p>c 4</p><div><p>d 5</p></div></div></div></div></div>";
    expect(renderToString(qweb, "Parent", { root })).toBe(expected);
    const recursiveFn = Object.values(qweb.subTemplates)[0] as any;
    expect(recursiveFn.toString()).toMatchSnapshot();
  });
```

The test's results are:
expected
```xml
<div><div><p>a 2</p><div><p>b 3</p><div><p>c 4</p><div><p>d 5</p></div></div></div></div></div>
```

Received
```xml
<div><div><p>a 2</p><div><p>b 2</p><div><p>c 2</p><div><p>d 2</p></div></div></div></div></div>
```

Long story short: setting a recursive index inside a template which, for each child, increments that index, OWL currently doesn't work

The issue is located mainly in the t-call scoping, where we do
```js
Object.assign(Object.create(context), scope).
```
Where `context` contains the initial recursive_idx (=2)
and `scope` doesn't contain recursive_idx in its first level properties.

ged-odoo:
actually fixed by referenced commit

PR #673, opened 2020-03-11
[FIX] qweb: protect scope in t-call writing recursively

kebeclibre:
Have something like
```xml
 <div t-name="Parent">
   <t t-call="nodeTemplate">
     <t t-set="recursive_idx" t-value="1"/>
     <t t-set="node" t-value="root"/>
   </t>
 </div>

 <div t-name="nodeTemplate">
   <t t-set="recursive_idx" t-value="recursive_idx + 1"/>
   <p><t t-esc="node.val"/> <t t-esc="recursive_idx"/></p>
   <t t-foreach="node.children or []" t-as="subtree">
     <t t-call="nodeTemplate">
       <t t-set="node" t-value="subtree"/>
     </t>
   </t>
 </div>
```

Where we want to propagate a recursion index through recursive t-calls

Before this commit, it did not work as we protected the scope in order
to not leak, in the wronf manner. Namely the protected scope only took
firt level prototype properties of the original scope.

After this commit, this case works as we mark the scope as read only

solves #672

### DIFF
--- a/src/qweb/compilation_context.ts
+++ b/src/qweb/compilation_context.ts
@@ -196,11 +196,12 @@ export class CompilationContext {
     const protectID = this.generateID();
     this.rootContext.protectedScopeNumber++;
     this.rootContext.shouldDefineScope = true;
-    const scopeExpr = codeBlock
-      ? `Object.create(scope);`
-      : `Object.assign(Object.create(context), scope);`;
+    const scopeExpr = `Object.create(scope);`;
     this.addLine(`let _origScope${protectID} = scope;`);
     this.addLine(`scope = ${scopeExpr}`);
+    if (!codeBlock) {
+      this.addLine(`scope.__access_mode__ = 'ro';`);
+    }
     return protectID;
   }
   stopProtectScope(protectID: number) {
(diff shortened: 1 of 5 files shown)


kebeclibre:
@ged-odoo 
I think the fix is not too bad, altough the paramer for getScope `codeBlock` now is not very expressive

src/qweb/compilation_context.ts
```diff
@@ -196,11 +196,12 @@ export class CompilationContext {
...
     this.rootContext.protectedScopeNumber++;
     this.rootContext.shouldDefineScope = true;
-    const scopeExpr = codeBlock
-      ? `Object.create(scope);`
-      : `Object.assign(Object.create(context), scope);`;
+    const scopeExpr = `Object.create(scope);`;
     this.addLine(`let _origScope${protectID} = scope;`);
     this.addLine(`scope = ${scopeExpr}`);
+    if (!codeBlock) {
+      this.addLine(`scope.__accmode__ = 'ro';`);
```
ged-odoo: `__access__mode__`

kebeclibre:
@ged-odoo 
Changes made

PR #689, opened 2020-04-20
[DOC] minor fixes

obayit:


doc/reference/animations.md
```diff
@@ -72,7 +72,7 @@ For example, a simple fade in/out effect can be done with this:
 
 ```xml
 <div>
-    <div t-if="state.flag" class="square" t-transition="fade">Hello</div>
+    <div class="square" t-transition="fade">Hello</div>
```
SimonGenin, line 75: It is better with the t-if. It emphases that there is a need of a switch between one state to another.

SimonGenin:
I will apply the `this.props` fix. 
However, not the other ones. The t-if feels relevant, and for the quick start file, it will likely get rewritten in the forseeable futur since owl evolved a lot.

SimonGenin:
https://github.com/odoo/owl/pull/784

PR #692, opened 2020-04-30
[IMP] hooks: improve useRef typings

ged-odoo:
With this commit, we can get the type for the target component for a
ref.

By default, if the generic type is not given, the ref will simply use
the base Component type.


aab-odoo:
ping @dmo-odoo

PR #697, opened 2020-05-14
[FIX] qweb: t-esc inside t-call using outside t-set

aab-odoo:
Before this commit, the generated code was incorrect, and crashed,
when there was a t-esc="abc" in a subtemplate, with t-set="abc"
done outside the subtemplate, with syntax <t t-set="abc">value</t>.

### DIFF
--- a/tests/qweb/qweb.test.ts
+++ b/tests/qweb/qweb.test.ts
@@ -207,6 +207,13 @@ describe("t-esc", () => {
     const result = renderToString(qweb, "test", { state: { list: [1, 2] } });
     expect(result).toBe("<span>1,2</span>");
   });
+
+  test("t-esc inside t-call, with t-set outside", () => {
+    qweb.addTemplate("main", `<div><t t-set="v">Hi</t><t t-call="sub"/></div>`);
+    qweb.addTemplate("sub", `<span t-esc="v"/>`);
+    const result = renderToString(qweb, "main");
+    expect(result).toBe("<div><span>Hi</span></div>");
+  });
 });
 
 describe("t-raw", () => {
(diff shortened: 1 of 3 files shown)


aab-odoo:
ping @ged-odoo @kebeclibre 
I could have called "shouldDefineUtils" above the if/else, once for both branches, but in this case it would be sometimes defined when it isn't necessary (when we are in a t-esc, and 'value.hasBody' is false).

tests/qweb/qweb.test.ts
```diff
@@ -207,6 +207,13 @@ describe("t-esc", () => {
     const result = renderToString(qweb, "test", { state: { list: [1, 2] } });
     expect(result).toBe("<span>1,2</span>");
   });
+
+  test("t-esc inside t-call, with t-set outside", () => {
+    qweb.addTemplate("main", `<div><t t-set="v">Hi</t><t t-call="sub"/></div>`);
+    qweb.addTemplate("sub", `<span t-esc="v"/>`);
+    const result = renderToString(qweb, "main");
```
kebeclibre, line 214: I am not sure this does test that the HTML is escaped ?
There is another test:
```js  
test("t-esc is escaped", () => {
    qweb.addTemplate("test", `<div><t t-set="var"><p>escaped</p></t><t t-esc="var"/></div>`);
    const domRendered = renderToDOM(qweb, "test");
    expect(domRendered.textContent).toBe("<p>escaped</p>");
  });
```

aab-odoo: Not sure it's relevant here as the value is hardcoded anyway (in this kind of scenario, you will either define html content, and use t-raw, or define string content and use t-set). I guess there are already a bunch of tests asserting that t-esc properly escapes the value, even inside a t-call, no?

kebeclibre: that would be my assumption too, yes, I was just making sure...

PR #716, opened 2020-07-07
[FIX] qweb: t-call should protect scope and let it accessible

### LINKED ISSUE #695: [BUG] t-foreach variables are lost in t-call inside a t-foreach inside itself
**JS**
```js
class AppCall extends owl.Component {}

// With t-call
const appCall = new AppCall();
appCall.mount(document.body);

class AppFlat extends owl.Component {}

// Without t-call
const appFlat = new AppFlat();
appFlat.mount(document.body);
```
**XML**
```xml
<templates>
  <!-- With t-call -->
  <t t-name="called">
    <t t-esc="a"/>
    <t t-esc="b"/>
    <t t-esc="c"/>
  </t>
  <t t-name="AppCall">
    <div>
      <t t-foreach="[1, 2, 3]" t-as="a" t-key="a">
        <t t-foreach="[4, 5, 6]" t-as="b" t-key="b">
          <t t-call="called">
            <t t-set="c" t-value="'x '"/>
          </t>
        </t>
      </t>
    </div>
  </t>
  <!-- Without t-call -->
  <t t-name="AppFlat">
    <div>
      <t t-foreach="[1, 2, 3]" t-as="a" t-key="a">
        <t t-foreach="[4, 5, 6]" t-as="b" t-key="b">
          <t t-set="c" t-value="'x '"/>
          <t t-esc="a"/>
          <t t-esc="b"/>
          <t t-esc="c"/>
        </t>
      </t>
    </div>
  </t>
</templates>
```
**Problem**
The results of the 2 templates are supposed to be the same.
-> Variables of the first `t-foreach` are lost in the scope of the called subtemplate

kebeclibre:
Have a t-call nested in a t-foreach nested in a t-foreach

```xml
<t t-name="template">
  <t t-foreach="..." t-as="a">
    <t t-foreach="..." t-as="b">
      <t-call="templateCalled" />
    </t>
  </t>
</t>
```

Before this commit, the `a` variable was not accessible within the t-call.
That was because the way t-call protected its scope by hiding other protected scope
in this case, the first protected scope for the first `t-foreach` was hidden

After this commit, `a` and `b` are accessible in the t-call, whether the t-call
defines its own variables by `t-set` or not.
Also, as expected from other fixes, there is no leaks of variables defined within a `t-call`

fixes #695

### DIFF
--- a/src/qweb/base_directives.ts
+++ b/src/qweb/base_directives.ts
@@ -245,12 +245,11 @@ QWeb.addDirective({
     // Step 3: compile t-call body if necessary
     // ------------------------------------------------
     let hasBody = node.hasChildNodes();
-    let protectID;
+    const protectID = ctx.startProtectScope();
     if (hasBody) {
       // we add a sub scope to protect the ambient scope
       ctx.addLine(`{`);
       ctx.indent();
-      protectID = ctx.startProtectScope();
       const nodeCopy = node.cloneNode(true) as Element;
       for (let attr of ["t-if", "t-else", "t-elif", "t-call"]) {
         nodeCopy.removeAttribute(attr);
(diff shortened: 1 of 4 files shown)


aab-odoo:
@ged-odoo

PR #724, opened 2020-08-31
[FIX] component: properly clean child vnode on destroy

seb-odoo:
When a child component is destroyed, the method does not clean its vnode from
its parent's children list, because it is assumed for performance that the parent
will be destroyed as well.

However, if the parent is not destroyed, it will be left with obsolete elements
in its children list. When that is the case, if a following render is bringing
back a previously deleted child, this child will not be added back in the DOM
due to the parent thinking it is already there.

### DIFF
--- a/src/component/component.ts
+++ b/src/component/component.ts
@@ -402,7 +402,7 @@ export class Component<Props extends {} = any, T extends Env = Env> {
    * Destroy the component.  This operation is quite complex:
    *  - it recursively destroy all children
    *  - call the willUnmount hooks if necessary
-   *  - remove the dom node from the dom
+   *  - remove the dom node from the dom and from the vdom of the parent
    *
    * This should only be called manually if you created the component.  Most
    * components will be automatically destroyed.
@@ -411,8 +411,18 @@ export class Component<Props extends {} = any, T extends Env = Env> {
     const __owl__ = this.__owl__;
     if (!__owl__.isDestroyed) {
       const el = this.el;
-      this.__destroy(__owl__.parent);
+      const parent = __owl__.parent;
+      this.__destroy(parent);
       if (el) {
+        if (parent && parent.__owl__.vnode && parent.__owl__.vnode.children) {
+          const childIndex = parent.__owl__.vnode.children.findIndex((n) => {
+            if (!n || typeof n === "string") {
+              return false;
+            }
+            return n.elm === el;
+          });
+          delete parent.__owl__.vnode.children[childIndex];
+        }
         el.remove();
       }
     }
--- a/src/core/observer.ts
+++ b/src/core/observer.ts
@@ -56,7 +56,7 @@ export class Observer {
           }
           self._updateRevNumber(target);
           target[key] = newVal;
-          self.notifyCB();
+          Promise.resolve(self.notifyCB()).catch(() => {});
         }
         return true;
       },
@@ -64,7 +64,7 @@ export class Observer {
         if (key in target) {
           delete target[key];
           self._updateRevNumber(target);
-          self.notifyCB();
+          Promise.resolve(self.notifyCB()).catch(() => {});
         }
         return true;
       },
--- a/tests/component/component.test.ts
+++ b/tests/component/component.test.ts
@@ -1104,6 +1104,29 @@ describe("destroy method", () => {
     expect(fixture.innerHTML).toBe("<div></div>");
     expect(steps).toEqual(["willStart", "__destroy"]);
   });
+
+  test("destroying a child should not prevent from mounting it again afterwards", async () => {
+    class MessageList extends Component {
+      static template = xml`<ul/>`;
+    }
+    class ThreadView extends Component {
+      static components = { MessageList };
+      static template = xml`<div><t t-if="state.hasMessages"><MessageList/></t></div>`;
+      state = useState({ hasMessages: true });
+    }
+    const threadView = new ThreadView();
+    await threadView.mount(fixture);
+    expect(fixture.innerHTML).toBe("<div><ul></ul></div>");
+
+    const fixture2 = makeTestFixture();
+    threadView.state.hasMessages = false;
+    await threadView.mount(fixture2); // TODO crash here...
+    expect(fixture2.innerHTML).toBe("<div></div>");
+
+    threadView.state.hasMessages = true;
+    await nextTick();
+    expect(fixture2.innerHTML).toBe("<div><ul></ul></div>");
+  });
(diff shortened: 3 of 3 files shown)


ged-odoo:
hmm, I don't want to be that guy, but do you have a test for your issue?

seb-odoo:
I preferred to share the issue before looking further, because it already took me long to even figure out what was the issue in the first place. (a very specific "scenario" in Odoo is required to have the issue)

I can always look longer to hopefully (?) craft a test that reproduces it, but I wished maybe issue would appear obvious/acceptable enough, because I'm not sure to get a test in a reasonable time.

ged-odoo:
you know that without a test, this will be difficult! I know that you are perfectly rigorous... 

Also, this commit may have a significant impact on performance when unmounting large lists of components. I fear that your are biased toward this because this is the solution you found, but there may be another way to solve your issue that has not this performance impact.  But to be able to discuss it, I need more information.

seb-odoo:
Ok, I'll spend the rest of today to try to get a test working.

If I can't craft a test it I'll explain a step by step what I think is the issue, and how to reproduce it in Odoo, then we can see together if need be.

seb-odoo:
I pushed a new version where the fix is moved into the public `destroy`.

I tried to make a test (the most naive version first) but I'm getting another error "Mounting operation cancelled" when I'm trying to successively update `state` and `mount` component to another target (see test currently in PR and line with TODO).

kebeclibre:
@seb-odoo 
I think I gotcha.

```js
        QUnit.test('change far child instance with outdated vnode', async function (assert) {
            assert.expect();

            class Custom extends Component {}
            Custom.template = owl.tags.xml`
                <div class="widget-subkey">
                    <t t-esc="props.key"/>__<t t-esc="props.subKey"/>
                </div>
            `;

            class Child extends Component {}
            Child.template = owl.tags.xml`
                <t t-component="Custom" key="props.key" subKey="props.subKey" t-key="props.subKey" />
            `;
            Child.components = { Custom };
                constructor() {
                    super(...arguments);
                    this.childProps = {
                        key: 1,
                        subKey: 1,
                    };
                }
            }
            Parent.template = owl.tags.xml`
                <div>
                    <div class="godfather">
                        <Child t-key="childProps.key" t-props="childProps"/>
                    </div>
                </div>
            `;

            const target = testUtils.prepareTarget();
            const parent = new Parent(null);
            await parent.mount(target);
            assert.strictEqual(parent.el.textContent.trim(), '1__1');

            // First step: change the Adapter's instance
            Object.assign(parent.childProps, {
                subKey: 2,
            });
            parent.render();
            await nextTick();
            assert.strictEqual(parent.el.textContent.trim(), '1__2');
            
            // Second step, change both Child's and Adapter's instance
            // crashes at insertBefore in Vdom
            Object.assign(parent.childProps, {
                key: 2,
                subKey: 3,
            });
            parent.render();
            await nextTick();
            assert.strictEqual(parent.el.textContent.trim(), '2__3');
        });
```

seb-odoo:
Thanks for sharing, as we checked together it is apparently and unfortunately not the same.

seb-odoo:
Just to give you a quick update on this.

We decided to use a workaround in our test that had the issue (dummy `await` to prevent the race condition...), so this issue becomes "less critical" for us, and I will not spend more time on it, at least for now.

Feel free to handle it as you see fit, and with the timing that you have available.
Note that I still think there is a real issue here that should definitely be handled at some point, sooner than later.

seb-odoo:
Of course I'm still available to help describe it or reproduce it if needed.

kebeclibre:
With the test:
```js
  test.only("destroying a sub-component cleans itself from parent's vnode", async () => {
    let c1;
    class C1 extends Component {
      static template = xml`<div><div><t t-esc="props.a"/></div></div>`;
      constructor() {
        super(...arguments);
        if (!c1) {
          c1 = this;
        }
      }
    }
    class P extends Component {
      static components = { C1 };
      static template = xml`<div><div><C1 t-props="state"/></div></div>`;
      state = {
        a: 1,
      }
    }
    const parent = new P();
    await parent.mount(fixture);
    expect(fixture.textContent).toBe('1');
    parent.state.a = 2;
    c1.destroy();
    await parent.render();
    expect(fixture.textContent).toBe('2');
  });
});
```

It is a **much** simpler use case than the real one (in the real one we must pass into `__render` L671 -- where children are destroyed when their parentLastFiber is outdated); but the results are the same. Also, the case is much harder to reach in reality -- I did not succeed doing that just yet.

When re-rendering the parent, a new component, with the SAME key and sel as the one before is created parent tries to patch its known children with it. Since the nodes have the same identification, the new won't replace the old.
Moreover, we expect old to still be in DOM, to get its parentElement.

This makes me think something is missing, either in component, or in vdom because:
- we use public destroy(), that can be expected to be used by anyone. The test case is then justified anyway.
- we can't be sure destroy is called *after* the old element has been replaced (this is what the vnode hook destroy is for)
- in destroy we could walk down the parent children tree to spot the children where to extract ourselves from (this may prove heavy for big components)
- in the directive we could push a hook on the pvnode to do just that (without crawling, since we already know what is the real vnode to be removed from)
- VDOM may do this, but its API should change (we need to pass the parent vnode, alter sameVnode, createElement at some point if vnode.elm is not in dom)

@ged-odoo @seb-odoo what do you think about this ?

cc @aab-odoo

seb-odoo:
Thanks for looking into this, it seems like you did a good investigation.

I honestly don't understand much of what you said, I never went that far into the low level code of OWL.

But hopefully that should be enough for @aab-odoo or @ged-odoo to guide you when they have some time to look at it.

kebeclibre:
Real test:

```js
  test.only("destroying a sub-component cleans itself from parent's vnode", async () => {
    let resolver;
    let prom;
    class C1 extends Component {
      static template = xml`<div><div><t t-esc="props.a"/></div></div>`;
      async willStart() {
        return prom;
      }
    }
    class P extends Component {
      static components = { C1 };
      static template = xml`<div><div><C1 t-props="state" t-if="state.a"/></div></div>`;
      state = {
        a: 'first',
      }
    }
    const parent = new P();
    await parent.mount(fixture);
    expect(fixture.textContent).toBe('first');
    parent.unmount();
    prom = new Promise(resolve => { resolver = resolve }) as any;
    parent.state.a = '';
    parent.mount(fixture);
    parent.state.a = 'fixed';
    resolver();
    await parent.render();
    expect(fixture.textContent).toBe('fixed');
  });
```

PR #726, opened 2020-09-01
[FIX] component, fiber: update props with virtual node should not crash

kebeclibre:
Have a hierarchy of A, B, C components where:

```xml
<div t-name="A">
  <div>
    <B t-key="key1"/>
  </div>
</div>

<t t-name="B">
  <C t-key="key2"/>
</t>

<div t-name="C">
  <div><t t-esc="keys_as_props" /></div>
</div>
```

The subtility of the issues lies in B, which doesn't have its own
concrete DOM element, rather, it borrows it from C.

With the sequence of events:
- change key2
C1 is destroyed and replaced by another instance, and another node.
B1 has its props updated and is patched with the C2's node (CRITICAL)
A1 is patched

- change key1 AND key2
C2 is destroyed
B1 is destroyed
A1 is patched replacing B1 by B2, and their nodes too (which at this point should be C2's to C3's)

Before this commit, at the CRITICAL point, the node representing the component itself
(technically its pvnode) was not updated with the new concrete node provided by B1 patch with C2 node
i.e. it held the previous node still
The second array of steps crashed because at A1 patch, the new B2 node would replace B1, which
was out of the DOM (removed because C1 was destroyed long before),
and therefore without a viable parent to insert B2 node.

After this commit, we update the component's pvnode after the patch which elm had possibly changed
There is no crash anymore for this use case.

### DIFF
--- a/src/component/fiber.ts
+++ b/src/component/fiber.ts
@@ -238,6 +238,12 @@ export class Fiber {
       } else {
         if (fiber.shouldPatch) {
           component.__patch(component.__owl__.vnode!, fiber.vnode!);
+          // When updating a Component's props (in directive),
+          // the component has a pvnode AND should be patched.
+          // However, its pvnode.elm may have changed if it is a High Order Component
+          if (component.__owl__.pvnode) {
+            component.__owl__.pvnode.elm = component.__owl__.vnode!.elm;
+          }
         } else {
           component.__patch(document.createElement(fiber.vnode!.sel!), fiber.vnode!);
           component.__owl__.pvnode!.elm = component.__owl__.vnode!.elm;
--- a/tests/component/component.test.ts
+++ b/tests/component/component.test.ts
@@ -3015,6 +3015,55 @@ describe("random stuff/miscellaneous", () => {
     expect(env.qweb.templates[Parent.template].fn.toString()).toMatchSnapshot();
     expect(fixture.innerHTML).toBe("<div><span>42</span></div>");
   });
+
+  test("update props of component without concrete own node", async () => {
+    class Custom extends Component {
+      static template = xml`
+        <div class="widget-subkey">
+          <t t-esc="props.key"/>__<t t-esc="props.subKey"/>
+        </div>`
+    }
+    class Child extends Component {
+      static components = { Custom };
+      static template = xml`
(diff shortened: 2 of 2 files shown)


src/component/fiber.ts
```diff
@@ -237,7 +237,15 @@ export class Fiber {
         component.__patch(target!, fiber.vnode!);
       } else {
         if (fiber.shouldPatch) {
+          const _elm = component.__owl__.vnode!.elm;
           component.__patch(component.__owl__.vnode!, fiber.vnode!);
+          // When updating a Component's props (in directive),
+          // the component has a pvnode AND should be patched.
+          // However, its pvnode.elm may have changed if it is "virtual"
+          // i.e. coming from the child.
```
aab-odoo, line 245: I don't understand the last part of last sentence

src/component/fiber.ts
```diff
@@ -237,7 +237,15 @@ export class Fiber {
...
+          if (component.__owl__.pvnode && _elm !== component.__owl__.vnode!.elm) {
```
aab-odoo, line 246: can't you get rid of the second part of the condition (and always update the pvnode's elm) ?

aab-odoo (commented):
Not sure to fully understand what's happening, but the fix might be ok. @ged-odoo will need to validate this anyway.

src/component/fiber.ts
kebeclibre: yes, I suppose I can, but there is a catch. I executed many times the test full test suite and on my branch, one test seem to crash undeterministicly. The second part of the condition seems to avoid that

tests/component/component.test.ts
```diff
@@ -3015,6 +3015,57 @@ describe("random stuff/miscellaneous", () => {
...
+          t-key="props.subKey"
+          key="props.key"
+          subKey="props.subKey"/>`
+    }
+
+    class Parent extends Component {
+      static components = { Child };
+      static template = xml`
+        <div>
+            <div class="godfather">
```
aab-odoo, line 3039: why?

kebeclibre: No real reason, I think it was to debug

ged-odoo (approved):

PR #735, opened 2020-09-16
[FIX] component: allow var with body as (textual) props

### LINKED ISSUE #670: Potentially wrong interpretation of t-set values when passed as props.
Easily reproduced in the playground, load `Components` example and replace xml with following :
```xml
<templates>
  <div t-name="Greeter" class="greeter" t-on-click="toggle">
    <t t-esc="state.word"/>, <t t-esc="props.name"/>
  </div>

  <div t-name="App">
    <t t-set="name">BLAH</t>
    <Greeter name="name"/>
  </div>
</templates>
```

"Hello, [object Object]" is displayed but I would expect to see "Hello, BLAH", am I wrong ?

(cc @seb-odoo @alexkuhn )

ged-odoo:
Before this commit, a variable set in a template with t-set could be
passed to a subcomponent as a prop only if it was only textual. If the
value was defined in the body of the t-set directive, it was passed, but
as a VDomArray, which means that it will only be displayed as [object
object] or something.

Also, we do not want to leak the VDomArray abstraction in userland (code
written by people using owl).

This issue is actually quite a problem in practice, because values in a
templates are translated, but not in attributes.  Therefore, using a
t-set directive with a body text content is the proper way to have
translated values at runtime.

closes #670

### DIFF
--- a/src/qweb/qweb.ts
+++ b/src/qweb/qweb.ts
@@ -90,6 +90,26 @@ function isComponent(obj) {
   return obj && obj.hasOwnProperty("__owl__");
 }
 
+class VDomArray extends Array {
+  toString() {
+    return vDomToString(this);
+  }
+}
+
+function vDomToString(vdom: VNode[]): string {
+  return vdom
+    .map((vnode) => {
+      if (vnode.sel) {
+        const node = document.createElement(vnode.sel);
+        const result = patch(node, vnode);
+        return (<HTMLElement>result.elm).outerHTML;
+      } else {
+        return vnode.text;
+      }
+    })
+    .join("");
+}
+
 const UTILS: Utils = {
   zero: Symbol("zero"),
   toObj(expr) {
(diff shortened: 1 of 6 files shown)


src/qweb/qweb.ts
```diff
@@ -90,6 +90,26 @@ function isComponent(obj) {
   return obj && obj.hasOwnProperty("__owl__");
 }
 
+class VDomArray extends Array {}
+
+VDomArray.prototype.toString = function () {
+  return vDomToString(this);
+};
```
aab-odoo, line 97: Is there a reason not to do
```ts
class VDomArray extends Array {
  toString() {
    return VDomToString(this);
  }
}
```
?

aab-odoo (approved):
Nice that you can revert the other fix :-)

aab-odoo (approved):

PR #736, opened 2020-09-16
[FIX] test: timing issue

### LINKED ISSUE #729: testing timing issue
In master, in some rare (?? it happens on LPE machine) cases, there is a non deterministic fail on a test:
```
PASS  tests/core/event_bus.test.ts

Summary of all failing tests
 FAIL  tests/hooks.test.ts
  ● hooks › can use onWillStart, onWillUpdateProps

    expect(received).toBe(expected) // Object.is equality

    Expected: "<div><span>2</span></div>"
    Received: "<div><span>1</span></div>"

      617 |     app.state.value = 2;
      618 |     await nextTick();
    > 619 |     expect(fixture.innerHTML).toBe("<div><span>2</span></div>");
          |                               ^
      620 |     expect(steps).toEqual([
      621 |       "on2ndStart",
      622 |       "slow",

      at Object.test (tests/hooks.test.ts:619:31)
```

If the `await nextTick()` is doubled, then the test pass, so this looks like owl is doing properly its job, but the way we coordinate the timing in the tests is incorrect. 

Now, we actually do something not really correct in the test helpers:
```ts
scheduler.requestAnimationFrame = function (callback: FrameRequestCallback) {
  setTimeout(callback, 1);
  return 1;
};
```
Removing this solves the issue, which is cool, but sadly, another test fail (determinstically):
```
 FAIL  tests/animations.test.ts
  ● animations › transitionInsert is called the correct amount of times

    expect(received).toBe(expected) // Object.is equality

    Expected: "<div><span class=\"\">blue</span></div>"
    Received: "<div><span class=\"chimay-enter chimay-enter-active\">blue</span></div>"

      423 |     await nextTick();
      424 |     widget.el!.querySelector("span")!.dispatchEvent(new Event("transitionend"));
    > 425 |     expect(fixture.innerHTML).toBe('<div><span class="">blue</span></div>');
          |                               ^
      426 |     expect(QWeb.utils.transitionInsert).toBeCalledTimes(1);
      427 | 
      428 |     widget.state.flag = false;

      at Object.<anonymous> (tests/animations.test.ts:425:31)
```
Maybe something is wrong with the transition test (it is actually quite fragile, since it tries to perform assertions mid-rendering).

So, we should probably do something like this:
1. remove the scheduler requestAnimationFrame override
2. Fix the transition test
3. rejoice

Hopefully, these 3 steps should lead us to a test environment which is deterministic across all environments.

ged-odoo:
Before this commit, we artificially replaced in the tests the
requestAnimationFrame by a setTimeout, to actually increase the speed of
the tests.  However, this is not really a true replacement.  For
example, a real setTimeout can come before or after a real
nextAnimationFrame, depending on when/where it is requested.

Also, this change exposed another problem: the nextTick function did a
setTimeout before a nextanimationframe. This is not a problem when
nextAnimationFrame is replaced by a setTimeout, because then all
expectations holds in owl.  However, it is wrong: to get to the next
animation frame, we need to request an animation frame, and THEN wait
with a setTimeout.

closes #729

### DIFF
--- a/tests/helpers.ts
+++ b/tests/helpers.ts
@@ -8,12 +8,6 @@ import "../src/qweb/extensions";
 import "../src/component/directive";
 import { browser } from "../src/browser";
 
-// modifies scheduler to make it faster to test components
-scheduler.requestAnimationFrame = function (callback: FrameRequestCallback) {
-  setTimeout(callback, 1);
-  return 1;
-};
-
 // Some static cleanup
 let nextSlotId;
 let slots;
@@ -43,9 +37,8 @@ export function nextMicroTick(): Promise<void> {
 }
 
 export async function nextTick(): Promise<void> {
-  return new Promise(function (resolve) {
-    setTimeout(() => scheduler.requestAnimationFrame(() => resolve()));
-  });
+  await new Promise((resolve) => scheduler.requestAnimationFrame(resolve));
+  await new Promise((resolve) => setTimeout(resolve));
 }
 
 export function makeTestFixture() {


tests/helpers.ts
```diff
@@ -43,9 +37,8 @@ export function nextMicroTick(): Promise<void> {
 }
 
 export async function nextTick(): Promise<void> {
-  return new Promise(function (resolve) {
-    setTimeout(() => scheduler.requestAnimationFrame(() => resolve()));
-  });
+  await new Promise((resolve) => scheduler.requestAnimationFrame(resolve));
+  await new Promise((resolve) => setTimeout(resolve));
```
aab-odoo, line 41: Interesting... That's not how it has been implemented in Odoo. In Odoo, we wait for a setTimeout before the nextAnimationFrame. I believed this was correct when we wrote it. We may need to talk about that to ensure that we have the correct implementation in both projects.

aab-odoo (approved):

PR #742, opened 2020-09-17
[IMP] qweb: add indeterminate to special input properties

### LINKED ISSUE #713: Add "indeterminate" to valid input[type=checkbox] props
**JS**
```js
    state = owl.useState({ indeterminate: true })
    cb = owl.hooks.useRef("cb")
    static template = owl.tags.xml`
        <div>
            <input type="checkbox" t-ref="cb" t-att-indeterminate="state.indeterminate"/>
        </div>`;
}

const app = new App();
app.mount(document.body).then(() => {
    console.log("Should be true:", app.cb.el.indeterminate)
});
```
**Problem**
The input `indeterminate` property yields false instead of true

ged-odoo:
Input with type="checkbox" have a special property (indeterminate) to
visually display the fact that the input value is non determinate (in my
chrome browser, the checkbox is then drawn with a simple - inside). It
does not actually modify the value of the input, only the way it is
displayed.

So, with this commit, owl will properly set the property, as expected.

closes #713

### DIFF
--- a/src/qweb/qweb.ts
+++ b/src/qweb/qweb.ts
@@ -66,8 +66,6 @@ interface QWebConfig {
 // Const/global stuff/helpers
 //------------------------------------------------------------------------------
 
-const DISABLED_TAGS = ["input", "textarea", "button", "select", "option", "optgroup"];
-
 const TRANSLATABLE_ATTRS = ["label", "title", "placeholder", "alt"];
 
 const lineBreakRE = /[\r\n]/;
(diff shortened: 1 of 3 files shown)


tests/qweb/qweb.test.ts
```diff
@@ -1895,6 +1895,14 @@ describe("special cases for some boolean html attributes/properties", () => {
     patch(vnode2, vnode3);
     expect(elm.value).toBe("potato");
   });
+
+  test("input of type checkbox with t-att-indeterminate", () => {
+    qweb.addTemplate("test", `<input type="checkbox" t-att-indeterminate="v"/>`);
```
aab-odoo, line 1900: what if we have both t-att-indeterminate and t-att-checked?

src/qweb/qweb.ts
```diff
@@ -676,24 +676,23 @@ export class QWeb extends EventBus {
...
-        let type = (<Element>node).getAttribute("type");
-        if (type === "checkbox" || type === "radio") {
+      if (node.nodeName === "input") {
+        if (key === "checked") {
+          let type = (<Element>node).getAttribute("type");
+          if (type === "checkbox" || type === "radio") {
+            isProp = true;
+          }
+        }
+        if (key === "value" || key === "readonly" || key === "indeterminate") {
```
aab-odoo, line 686: indeterminate not only for checkboxes and radio (for tagName input)? Also, it seems to be a valid attribute on `<progress>` tag as well.

src/qweb/qweb.ts
```diff
@@ -676,24 +676,23 @@ export class QWeb extends EventBus {
...
-      if (node.nodeName === "input" && key === "checked") {
-        let type = (<Element>node).getAttribute("type");
-        if (type === "checkbox" || type === "radio") {
+      if (node.nodeName === "input") {
+        if (key === "checked") {
+          let type = (<Element>node).getAttribute("type");
+          if (type === "checkbox" || type === "radio") {
+            isProp = true;
+          }
+        }
```
aab-odoo, line 685:
```      if (node.nodeName === "input") {
              let type = (<Element>node).getAttribute("type");
        if (type === "checkbox" || type === "radio") {
          if (key === "checked" || key === "indeterminate") {
              isProp = true;
          }
        }
```

aab-odoo, line 685:
```suggestion
     let type = (<Element>node).getAttribute("type");
        if (type === "checkbox" || type === "radio") {
          if (key === "checked" || key === "indeterminate") {
              isProp = true;
          }
        }
```

aab-odoo (approved):

PR #751, opened 2020-10-12
[IMP] bundling: Improve owl bundling process and output

SimonGenin:
The bundle pipeline can use a bit of a improvement:

Most of the process is moved from npm scripts to rollup.
Add package keys to smooth out the use of bundlers of the end users.
(Rollup by example check amongst others the main, module and browser key
to import the right version of a lib when needed).
Refactor rollup logic to make it more modular.
Add bundled module formats.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,22 +2,19 @@
   "name": "@odoo/owl",
   "version": "1.0.11",
   "description": "Odoo Web Library (OWL)",
-  "main": "dist/owl.js",
+  "main": "dist/owl.cjs.js",
+  "browser": "dist/owl.iife.js",
+  "module": "dist/owl.es.js",
   "types": "dist/types/index.d.ts",
   "files": [
-    "dist/types/",
-    "dist/owl.js",
-    "dist/owl-iife.js"
+    "dist"
   ],
   "engines": {
     "node": ">=10.15.3"
   },
   "scripts": {
-    "build:js": "tsc --target esnext --module es6 --outDir dist/owl",
     "build:bundle": "rollup -c",
-    "build": "npm run build:js && npm run build:bundle",
-    "buildcommonjs": "npm run build:js && npm run build:bundle -- -f cjs",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -2,30 +2,30 @@
...
+  "module": "dist/owl.es.js",
+  "system": "dist/owl.system.js",
+  "amd": "dist/owl.amd.js",
+  "umd": "dist/owl.umd.js",
   "types": "dist/types/index.d.ts",
   "files": [
-    "dist/types/",
-    "dist/owl.js",
-    "dist/owl-iife.js"
+    "dist"
```
ged-odoo, line 13: you should probably find a way to exclude `dist/js`

package.json
```diff
@@ -2,30 +2,30 @@
...
   "engines": {
     "node": ">=10.15.3"
   },
   "scripts": {
-    "build:js": "tsc --target esnext --module es6 --outDir dist/owl",
     "build:bundle": "rollup -c",
-    "build": "npm run build:js && npm run build:bundle",
-    "buildcommonjs": "npm run build:js && npm run build:bundle -- -f cjs",
-    "minify": "uglifyjs dist/owl.js -o dist/owl.min.js --compress --mangle",
+    "build": "npm run build:bundle",
```
ged-odoo, line 20: maybe remove the `build:bundle` task?

SimonGenin:
@ged-odoo I just trie the test workflow, it looks good, I didn't feel anything was slower with the watch :)

package.json
```diff
@@ -2,30 +2,30 @@
   "name": "@odoo/owl",
   "version": "1.0.11",
   "description": "Odoo Web Library (OWL)",
-  "main": "dist/owl.js",
+  "main": "dist/owl.cjs.js",
+  "browser": "dist/owl.iife.js",
```
ged-odoo, line 6: remove some of them as we discussed

ged-odoo (approved):

PR #754, opened 2020-10-13
[FIX/IMP]: t-foreach can interpret any iterable.

### LINKED ISSUE #720: t-foreach not supporting custom iterators
Due to
```ts
    ctx.addLine(`_${keysID} = Object.keys(_${arrayID});`);
    ctx.addLine(`_${valuesID} = Object.values(_${arrayID});`);
```
`t-foreach` is not supporting custom iterators. For example, but not limited to, using native `Set`.

Suggestion: replace the above lines by
```ts
ctx.addLine(`_${arrayID} = [..._${arrayID}];`);
```
This works for Set (or any value where `typeof _${arrayID}[Symbol.iterator] === 'function'`) but probably not for objects though. 
Alternative would be to use
```ts
`_${arrayID}.keys()`
`_${arrayID}.values()`
```
or
```ts
`_${arrayID}.constructor.keys(_${arrayID})`
`_${arrayID}.constructor.values(_${arrayID})`
```
for the objects that do have them.

SimonGenin:
t-foreach accepted only arrays and object.
Now, iterables and objects are accepted.

Closes #720

### DIFF
--- a/src/qweb/base_directives.ts
+++ b/src/qweb/base_directives.ts
@@ -312,8 +312,12 @@ QWeb.addDirective({
     let valuesID = ctx.generateID();
     ctx.addLine(`let _${keysID} = _${valuesID} = _${arrayID};`);
     ctx.addIf(`!(_${arrayID} instanceof Array)`);
+    ctx.addIf(`typeof _${arrayID}[Symbol.iterator] === 'function'`);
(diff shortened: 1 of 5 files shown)


SimonGenin:
Aaron explained me the snapshots so all good

PR #755, opened 2020-10-13
Create deploy.yml

SimonGenin:
Add action to launch the tests on the PRs


ged-odoo:
very cool... except that it does not work (we cannot use github actions because of pricing issue) :(

ged-odoo (approved):

PR #758, opened 2020-10-16
[IMP] component: allow using TypeScript type hints

brboi:
Just to have nice and beautiful type hinting like that:
image

### DIFF
--- a/doc/reference/event_handling.md
+++ b/doc/reference/event_handling.md
@@ -78,6 +78,19 @@ The `t-on` directive allows to prebind its arguments. For example,
 Here, `expr` is a valid Owl expression, so it could be `true` or some variable
 from the rendering context.
 
+### Type Hinting
+
+Note that if you work with Typescript, the `trigger` method is generic on the type of the payload.
+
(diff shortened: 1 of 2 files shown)


doc/reference/event_handling.md
```diff
@@ -78,6 +78,15 @@ The `t-on` directive allows to prebind its arguments. For example,
 Here, `expr` is a valid Owl expression, so it could be `true` or some variable
 from the rendering context.
 
+### Type Hinting
+You could also indicate your payload type to enforce your typings. You will see typing errors while developing and resolve it even before it's deployed.
```
ged-odoo, line 82: this section only applies for people working on typescript. This should be explained clearly, otherwise some people will complain that the code from this example does not work in JS.

Maybe add a short sentence to explain, at the beginning of this paragraph: 
```
Note that if you work with Typescript, the `trigger` method is generic on the type of the payload. You can then describe the type of the event, so you will see typing errors...
```

brboi: Done ;-) What do you think ?

ged-odoo (approved):

PR #771, opened 2020-10-20
Improve the release script for easier deployment

SimonGenin:
@ged-odoo you'll need to install https://cli.github.com/. 
The first time you will use it, it will ask you to launch the command gh login to connect to your github action.
Maybe do it first to avoid problems. 

I'd like a rollback feature if the last steps of the deployment fail. Because we have the problem of a new commit and a push being done. But because of the push, we would need a --force so it's not great.

### DIFF
--- a/tools/release.js
+++ b/tools/release.js
@@ -3,10 +3,9 @@ const readline = require("readline");
 const fs = require("fs");
 const exec = require("child_process").exec;
 const chalk = require("chalk");
-const GitHub = require("github-api");
 
 const REL_NOTES_FILE = `release-notes.md`;
-const STEPS = 10;
+const STEPS = 8;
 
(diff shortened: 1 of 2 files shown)


tools/release.js
```diff
@@ -68,61 +67,39 @@ async function startRelease() {
...
+    log("github release failed. Aborting.");
+    return;
+  }
 
-  // ---------------------------------------------------------------------------
-  log(`Step 10/${STEPS}: publishing module on npm...`);
+  log(`Step 8/${STEPS}: publishing module on npm...`);
   await execCommand("npm run publish");
-
+
```
aab-odoo, line 98: npm run prettier!!!

ged-odoo (approved):

PR #775, opened 2020-10-23
[FIX] slots: prevent infinite loop in some cases

ged-odoo:
The reviewer for the commit on named slots inside named slots did not
notice that there was an infinite loop.  Because of his sloppiness, Owl
could block in an infinite loop when a named t-slots was defined inside
a subcomponent, but not as a direct child.


aab-odoo (approved):
I approve even tough I don't like the commit message

PR #785, opened 2020-11-03
Create cla.yml

SimonGenin:
Add a github action to check sure the CLA of odoo has been signed.

How does it work: 
When a PR is created, the action check if the user has either signed the CLA or the user is in a whitelist (a few github names from the js team + every user name ending with "-odoo" + all the bots). 
If it's not the case, a comment will ask the user to add a comment looking like "I agree with the cla bla bla bla". When this comment is detected, a small commit is made to the project adding his signature inside the signature/cla.json file. 
There is also the possibility to relaunch the cla action by adding a comment "cla check". Useful if we add a person to the whitelist and want to make the action green after that.

There are a few options possible: original doc here : https://github.com/cla-assistant/github-action

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -1,0 +1,32 @@
+name: "CLA Assistant"
+on:
+  issue_comment:
+    types: [created]
+  pull_request_target:
+    types: [opened,closed,synchronize]
+    
+jobs:
+  CLAssistant:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


SimonGenin:
@ged-odoo There is the need of a secret token. Right now it comes from my profile. I am not sure it will work since I'm not the owner. But anyway, we should look to add it from your profile as you're the main contributor.

aab-odoo:
Closed for now, we'll reopen it later if we need it

PR #787, opened 2020-11-09
[IMP] devtools: add __owl__ access.

SimonGenin:
To be able to make devtools, we need access to the __owl__ variable
on a component. It contains most of the information a devtool may
required.

Just like in Vue js, we attach it to the HTMLElement of a component.

For its update, we use the method patched. On each patched call, a new
__owl__ is provided.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
     "node": ">=10.15.3"
   },
   "scripts": {
-    "build:bundle": "rollup -c",
-    "build": "npm run build:bundle",
+    "dev": "rollup -c",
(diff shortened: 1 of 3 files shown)


SimonGenin:
@ged-odoo I'm interested in your opinion. Here's where I took inspiration: https://github.com/vuejs/vue/search?q=__vue__

SimonGenin:
I improved it a bit already with the help of Lucas.

SimonGenin:
/!\ It added a circular dependency. Needs to be fixed somehow.

PR #789, opened 2020-11-12
forward data-* attributes to t-component root node

brboi:


src/component/directive.ts
```diff
@@ -271,13 +271,39 @@ QWeb.addDirective({
     let tattClass = node.getAttribute("t-att-class");
     let styleAttr = node.getAttribute("style");
     let tattStyle = node.getAttribute("t-att-style");
+
+    const dataAttrs = [...node.attributes].filter(a => /^(t-attf?-)?data-.+$/.test(a.name)).sort();
+    if (dataAttrs.length) {
+      const dashToCamel = (str: string) => {
```
aab-odoo, line 277: dash style is called kebab-case, so maybe you want to rename into `kebabToCamel` ?

src/component/directive.ts
```diff
@@ -271,13 +271,39 @@ QWeb.addDirective({
     let tattClass = node.getAttribute("t-att-class");
     let styleAttr = node.getAttribute("style");
     let tattStyle = node.getAttribute("t-att-style");
+
+    const dataAttrs = [...node.attributes].filter(a => /^(t-attf?-)?data-.+$/.test(a.name)).sort();
```
aab-odoo, line 275: sort?

src/component/directive.ts
```diff
@@ -271,13 +271,39 @@ QWeb.addDirective({
...
+    if (dataAttrs.length) {
+      const dashToCamel = (str: string) => {
+        return str.toLowerCase()
+          .replace(/(\-[a-z])/g, val => val.toUpperCase().replace('-', ''));
+      };
+      ctx.addLine("let dataAttrMap = {};")
```
aab-odoo, line 281: could be `const`

src/component/directive.ts
```diff
@@ -271,13 +271,39 @@ QWeb.addDirective({
...
+      let attrName = "", attrValue = "";
```
aab-odoo, line 282: typically, we try not to do that, and rather define one variable per line (full definition: `let machin = "val";`)

src/component/directive.ts
```diff
@@ -271,13 +271,39 @@ QWeb.addDirective({
...
+      for (const dataAttr of dataAttrs) {
+        if (dataAttr.name.startsWith("t-att-")) {
+          attrName = dashToCamel(dataAttr.name.slice(11));
+          let v = ctx.getValue(dataAttr.value);
+          attrValue = typeof v === "string" ? ctx.formatExpression(v) : `scope.${v.id}`;
```
aab-odoo, line 287: could this not be a string?

src/component/directive.ts
```diff
@@ -271,13 +271,39 @@ QWeb.addDirective({
...
+        } else if (dataAttr.name.startsWith("t-attf-")) {
+          attrName = dashToCamel(dataAttr.name.slice(12));
+          let v = ctx.getValue(dataAttr.value);
+          attrValue = typeof v === "string" ? ctx.interpolate(v) : `scope.${v.id}`;
```
aab-odoo, line 291: same

src/component/directive.ts
```diff
@@ -271,13 +271,39 @@ QWeb.addDirective({
...
+        } else {
+          attrName = dashToCamel(dataAttr.name.slice(5));
+          attrValue = '"' + dataAttr.value + '"';
```
aab-odoo, line 294: maybe use template string

src/component/directive.ts
```diff
@@ -271,13 +271,39 @@ QWeb.addDirective({
...
+        }
+        ctx.addLine(`dataAttrMap.${attrName} = dataAttrMap.${attrName} ? dataAttrMap.${attrName} + " " + ${attrValue} : ${attrValue};`)
```
aab-odoo, line 296: Not sure it makes sense to combine (by concatenation) data values. It works well for classnames, but here I guess I wouldn't do that.

aab-odoo (commented):
As we discussed, it's up to @ged-odoo to decide whether or not we want this. But in any case, good effort because this code isn't easy to dive into.

src/component/directive.ts
brboi: This needs to be discussed.
What to do when template looks like this:
```xml
<Component t-att-data-foo="var" data-foo="42" t-attf-data-foo="var is {{var}}" />
```

Sorting is not the solution, though :-D

src/component/directive.ts
brboi: I need to triple check, but I think this could be a QWebVar too

src/component/directive.ts
brboi: I agree. We should probably log warnings when template define more than once the same data-* attribute.

src/component/directive.ts
brboi: You were right.

LuvForAirplanes:
This was fixed in #903. Thanks for the effort.

PR #792, opened 2020-11-20
[FIX/IMP] export browser object, prevent crash in private mode

### LINKED ISSUE #791: Owl not loading inside iframe in incognito mode
When trying to load owl inside an iframe in incognito mode, we are faced with the following error:
`DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`

which point to:
```javascript
const browser = {
        setTimeout: window.setTimeout.bind(window),
        clearTimeout: window.clearTimeout.bind(window),
        setInterval: window.setInterval.bind(window),
        clearInterval: window.clearInterval.bind(window),
        requestAnimationFrame: window.requestAnimationFrame.bind(window),
        random: Math.random,
        Date: window.Date,
        fetch: (window.fetch || (()=>{}
        )).bind(window),
        localStorage: window.localStorage,
    };
```
From the information found over the internet (src: [here](https://stackoverflow.com/questions/63195318/domexception-failed-to-read-the-localstorage-property-from-window-in-chrom)) the problem is related to the cookie from the request being not set (chrome enforced its sameSite cookie policy a few month ago). Therefore if the user did not authorize cookie in incognito (which is the default setting and you would probably don't want to change that) or if you can't change the cookie policy of the request (for security reason or ...). Owl won't load.
I've tested by authorizing cookies in incognito, this indeed fixes the problem, in my case the cookie that is being sent is a simple session cookie (server send them by default with every requests and we can't change that fact easily, also we don't use that cookie).

Adding some sort of fallback would probably be nice to prevent it from crashing

Version of owl used: 1.0.13

Setup to reproduce using Odoo framework:
create a route
```python
@http.route(['/iframe_faulty'], methods=['GET'], type='http', auth='public', save_session=False)
    def open_iframe(self, **data):
        response = request.env.ref('module_name.faulty_template')._render()
        return request.make_response(response, headers=[
            ('Cache-Control', 'no-cache'),
            ('Content-Type', 'text/html; charset=utf-8'),
            ('Access-Control-Allow-Origin', '*'),
            ('Access-Control-Allow-Methods', 'GET')
        ])
```
Create the template associated with that route (I've removed pretty much all assets except owl)
```xml
        <template id="module_name.faulty_template" name="template_faulty">
            <html lang="en">
            <head>
                <meta charset="UTF-8" />
                <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1"/>
                <title>Faulty</title>
                <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no"/>
                <script type="text/javascript" src="/web/static/lib/owl/owl.js"></script>
            </head>
            <body></body>
            </html>
        </template>
```
and now open any webpage in incognito and paste the following code in the console (change the url before hand):
```javascript
    let iframe = null;
    const url = 'http://yoururl/iframe_faulty';
    const closeIframe = function() {
        iframe.parentNode.removeChild(iframe);
    }
    const openIframe = function() {
        iframe = document.createElement('iframe');
        iframe.src = url
        document.body.appendChild(iframe);
    }
    openIframe()
```

ged-odoo:
This commit performs two tasks:

1. it exports the internal browser object (in owl.browser), for ease of
testing
2. it makes sure that Owl does not crash inside iframes in incognito
mode, because window.localStorage raises an exception in that case

closes #791

### DIFF
--- a/doc/reference/browser.md
+++ b/doc/reference/browser.md
@@ -1,0 +1,33 @@
+# 🦉 Browser 🦉
+
+## Content
+
+- [Overview](#overview)
+- [Browser Content](#browser-content)
+
+## Overview
+
+The browser object contains some browser native APIs, such as `setTimeout`, that
+are used by Owl and its utility functions. They are exposed with the intent of
+making them mockable if necessary.
+
+```js
+owl.browser.setTimeout === window.setTimeout; // return true
+```
+
+For now, this object contains some functions that are not used by Owl. They
+will eventually be removed in Owl 2.0.
+
+## Browser Content
+
+More specifically, the `browser` object contains the following methods and objects:
+
+- `setTimeout`
+- `clearTimeout`
+- `setInterval`
+- `clearInterval`
+- `requestAnimationFrame`
+- `random`
+- `Date`
+- `fetch`
+- `localStorage`
--- a/doc/readme.md
+++ b/doc/readme.md
@@ -17,6 +17,7 @@ You will find here a complete reference of every feature, class or object
 provided by Owl.
 
 - [Animations](reference/animations.md)
+- [Browser](reference/browser.md)
 - [Component](reference/component.md)
 - [Content](reference/content.md)
 - [Concurrency Model](reference/concurrency_model.md)
--- a/doc/reference/animations.md
+++ b/doc/reference/animations.md
@@ -52,7 +52,7 @@ sequence of events will happen:
 At node insertion:
 
 - the css classes `name-enter` and `name-enter-active` will be added directly
-  when the node is inserted into the DOM. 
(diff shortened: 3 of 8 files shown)


doc/reference/browser.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Content
+
+- [Overview](#overview)
+- [Browser Content](#browser-content)
+
+## Overview
+
+The browser object contains some browser native APIs, such as `setTimeout`, that
+are used by Owl and its utility function. They are exposed with the intent of
```
aab-odoo, line 11: functions?

aab-odoo (approved):

PR #806, opened 2020-12-11
[FIX] store: properly render, even if shouldupdate is implemented

### LINKED ISSUE #799: useState not rendering when also having useStore (useContextWithCB)
I'm not sure if the error always happens yet, but here are the steps to reproduce:

- Have a component with `useStore` and `useState`
- Change state
- Observe there is no render

This is due to this condition preventing the render:
```js
            if (ctx.rev > mapping[id]) {
                // in this case, the context has been updated since we were rendering
                // last, and we do not need to render here with the observer. A
                // rendering is coming anyway, with the correct props.
                return;
            }
```
Except that in the case I am in, this render is actually needed.

I think this might be due to the component sharing the same observer for the `useStore` and `useState`, and this logic is only supposed to apply to `useStore`, so `useState` does not work correctly anymore.

I will try to reproduce on playground and keep posting here, but if you read this and the current report is enough let me know so I don't spend too much time getting a simple example working.

ged-odoo:
Before this commit, an unwanted behaviour happened when using components
with shouldUpdate implemented, and store/state.

The actual problem is the following: the sub component is using a store,
and shouldupdate.  Whenever the component is rendered, it checks if there is an incoming
rendering from the context.  If that is the case, it skips the
rendering, because we actually only want to be rendered by the store
rendering (otherwise, we may run into issue with inconsistent data (more
recent data from the context, older data in the component).

However, if shouldUpdate is implemented, then the rendering coming from
the context simply does not arrive.  To fix this, we just force these
renderings to go through all children, regardless of their shouldUpdate
status.

closes #799

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -1,5 +1,4 @@
-import { Component } from "./component/component";
-import { Env } from "./component/component";
+import { Component, Env } from "./component/component";
 import { Context, useContextWithCB } from "./context";
 import { onWillUpdateProps } from "./hooks";
 
--- a/src/context.ts
+++ b/src/context.ts
@@ -115,16 +115,6 @@ export function useContextWithCB(ctx: Context, component: Component, method): an
     __owl__.observer = new Observer();
     __owl__.observer.notifyCB = component.render.bind(component);
   }
-  const currentCB = __owl__.observer.notifyCB;
-  __owl__.observer.notifyCB = function () {
-    if (ctx.rev > mapping[id]) {
-      // in this case, the context has been updated since we were rendering
-      // last, and we do not need to render here with the observer. A
-      // rendering is coming anyway, with the correct props.
-      return;
-    }
-    currentCB();
-  };
 
   mapping[id] = 0;
   const renderFn = __owl__.renderFn;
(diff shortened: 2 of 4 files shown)


ged-odoo:
maybe not merge this....

src/store.ts
```diff
@@ -123,7 +123,7 @@ export function useStore(selector, options: SelectorOptions = {}): any {
       shouldRender = fn() || shouldRender;
     }
     if (shouldRender) {
-      return component.render();
+      return component.render(true);
```
seb-odoo, line 126: As per discussed, what I would like in this case is to force the render of the current component when it's observed state changes (the current diff should be good for that), but at the same time I don't want children component to be force-rendered too. I want the children to be rendered only if that makes sense according to their own `shouldUpdate()` value, or if their own observed state changes.

seb-odoo: Btw just to be clear, I'm not saying children should never render in this context, just that their `shouldUpdate()` should be respected: no render if it returns `false`, but since the default value is `true`, they will still be rendered in normal cases.
This gives the responsibility to whoever is overriding that method to ensure it returns the proper value according to what needs to be done.

aab-odoo (approved):

PR #819, opened 2021-01-06
[FIX] component: propagate correct info when reusing fibers

### LINKED ISSUE #818: missing render with shouldUpdate false on reuse/remap fiber
### Suggested fix:
```js
                if (oldFiber.force) {
                    this.force = true;
                }
```
inside `if (oldFiber && !oldFiber.isCompleted) {`

### Playground test below:
(this might not be the "minimal" test, it probably has remaining crap from previous attempts, but at least it's wrong without fix and good after fix)

Bug result:

> parent: 2
> child:
> child child: 1

Expected result:

> parent: 2
> child:
> child child: 2

```js
const { Component, useState } = owl;
const { useRef, useStore, onWillUpdateProps } = owl.hooks;

const store = new owl.Store({ state: { p: 1, cc: 1 }});


class ChildChild extends Component {
    constructor() {
        super(...arguments);
        this.res = useStore((state, props) => {
            return state.cc;
        }, {
            onUpdate: res => {
                this.res = res;
            },
        });
    }
    shouldUpdate() {
        return false;   
    }
}

class Child extends Component {
    shouldUpdate() {
        return false;   
    }
}
Child.components = { ChildChild };

    constructor() {
        super(...arguments);
        this.res = useStore((state, props) => {
            return state.p;
        }, {
            onUpdate: res => {
                this.res = res;
            },
        });
        this.child = useRef('child');
    }
    shouldUpdate() {
        return false;   
    }
}

class App extends Component {}
App.components = { Parent };


(async function () {
    var div = document.createElement("div");
    document.body.appendChild(div);
    
    Parent.env.store = store;
    const app = new App();

    console.log('------------------- mount');
    await app.mount(document.body);
    await new Promise(setTimeout);


    console.log('------------------- re mount');
    app.mount(div);

    await Promise.resolve();
    console.log('------------------- update');
    store.state.cc++;
    store.state.p++;
    
})();
```

```xml
<templates>
  <div t-name="ChildChild">
    child child: <t t-esc="res"/>
  </div>

  <div t-name="Child">
    child: <t t-esc="res"/>
    <ChildChild/>
  </div>

  <div t-name="Parent">
    parent: <t t-esc="res"/>
    <Child t-ref="child"/>
  </div>
  <div t-name="App">
    <Parent/>
  </div>
</templates>

```

ged-odoo:
In some cases, a rendering initiated in some component is then remapped
into a larger rendering initiated by some parent.

If we have some components which implement shouldUpdate to return false,
then the following scenario can happen:

- some parent component is mounted (which triggers a rendering with
force: true => bypass the shouldUpdate)
- some sub component is updated and rerendered, AFTER the previous
rendering goes through it
- the sub component notices that there is an ongoing rendering, and
  remaps itself in the parent rendering

Before this commit, the new fiber in the subcomponent does not have
force flag set to true, so the new rendering for the subcomponent does
not go through its own children (if they have shouldUpdate=false)

With this commit, we make sure that the flag of the new fiber is set to
true.

closes #818

### DIFF
--- a/tests/component/async.test.ts
+++ b/tests/component/async.test.ts
@@ -1405,4 +1405,160 @@ describe("async rendering", () => {
     expect(fixture.innerHTML).toBe("<div>2</div>");
     expect(Widget.prototype.__render).toHaveBeenCalledTimes(2);
   });
+
+  test("components with shouldUpdate=false", async () => {
+    const state = { p: 1, cc: 10 };
+
+    class ChildChild extends Component {
+      static template = xml`
+        <div>
+          child child: <t t-esc="state.cc"/>
+        </div>`;
+      state = state;
+      shouldUpdate() {
+        return false;
+      }
+    }
+
+    class Child extends Component {
+      static components = { ChildChild };
+      static template = xml`
+        <div>
+          child
+          <ChildChild/>
+        </div>`;
+
+      shouldUpdate() {
+        return false;
+      }
+    }
+
+    let parent: any;
+    class Parent extends Component {
+      static components = { Child };
+      static template = xml`
+        <div>
+          parent: <t t-esc="state.p"/>
+          <Child/>
+        </div>`;
+
+      state = state;
+      constructor(a, b) {
+        super(a, b);
+        parent = this;
+      }
+      shouldUpdate() {
+        return false;
+      }
+    }
+
+    class App extends Component {
+      static components = { Parent };
+      static template = xml`
+        <div>
+          <Parent/>
+        </div>`;
+    }
+
+    var div = document.createElement("div");
+    fixture.appendChild(div);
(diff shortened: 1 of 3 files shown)


tests/component/async.test.ts
```diff
@@ -1405,4 +1405,160 @@ describe("async rendering", () => {
...
+    );
+
+    state.cc++;
+    state.p++;
+    app.render();
+
+    // wait for rendering to go through parent
+    await Promise.resolve();
+    await Promise.resolve();
+    await Promise.resolve();
```
aab-odoo, line 1555: 2 microticks were enough in the previous test, what changed?

aab-odoo (approved):

PR #820, opened 2021-01-08
[FIX] store: properly call onUpdate functions in some cases

### LINKED ISSUE #816: useStore ignored during remount
I stumbled upon this randomly while trying to reproduce another bug.

I'm expecting the result to be

> parent: 1
> child: 2
> child child 

But I am getting

> parent: 1
> child: 1
> child child 

Awaiting/removing the last mount shows the desired behavior. Somehow this store update is ignored if it happens just as the mount is in progress (from what I can tell, the problematic update happens after the fiber of Child is created but before the one from ChildChild is created).

```js
const { Component, useState } = owl;
const { useStore, onWillUpdateProps } = owl.hooks;

const store = new owl.Store({ state: { a: 1, b: 1 }});


class ChildChild extends Component {
    constructor() {
        super(...arguments);
    }
    shouldUpdate() {
        return false;
    }

}

class Child extends Component {
    constructor() {
        super(...arguments);
        this.a = useStore((state, props) => {
            return state.a;
        }, {
            onUpdate: () => {
                this.a = store.state.a;
            },
        });
    }
    shouldUpdate() {
        return false;
    }
}
Child.components = { ChildChild };

    constructor() {
        super(...arguments);
        this.b = useStore((state, props) => {
            return state.b;
        }, {
            onUpdate: () => {
                this.b = store.state.b;
            },
        });
    }
    shouldUpdate() {
        return false;
    }
}

(async function () {
    var div = document.createElement("div");
    document.body.appendChild(div);
    
    Parent.env.store = store;
    const app = new Parent();
    await app.mount(document.body);
    console.log('------------------- remount');
    app.mount(div); // await or remove this line to see "child: 2"
    console.log('------------------- change a');
    store.state.a++;
})();
```

```xml
<templates>
  <div t-name="ChildChild">
    child child
  </div>

  <div t-name="Child">
    child: <t t-esc="a"/>
    <ChildChild/>
  </div>

  <div t-name="Parent">
    parent: <t t-esc="b"/>
    <Child/>
  </div>
</templates>
```

ged-odoo:
Before this commit, the following scenario could happen:

Suppose that we have a parent component A, connected to a store,
 and a child component B, also connected to the store and using
the onUpdate feature.

Then, we remount the A component in some other places and a
rendering is initiated in A.  We immediately update the store state.
What happens next is:

- rendering A is done, A internal revid is updated
- store update A (but nothing is done because the state change here
  does not modify A)
- rendering B is done (from parent), B internal revid is updated
- store update B, notice internal revid is updated, does not call the
  onUpdate function

We then have the B component which has not its internal state updated,
because we did not call its onUpdate function.

The solution is to move the onUpdate call in a "preupdate" event, to be
sure that it is called everytime the store is updated.

closes #816

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -75,6 +75,11 @@ export class Store extends Context {
     );
     return result;
   }
+
+  __notifyComponents(): Promise<void> {
+    this.trigger("before-update");
+    return super.__notifyComponents();
+  }
 }
 
 interface SelectorOptions {
@@ -105,12 +110,15 @@ export function useStore(selector, options: SelectorOptions = {}): any {
     const newRevNumber = hashFn(result);
     if ((newRevNumber > 0 && revNumber !== newRevNumber) || !isEqual(oldResult, result)) {
       revNumber = newRevNumber;
-      if (options.onUpdate) {
-        options.onUpdate(result);
-      }
       return true;
     }
     return false;
+  }
+  if (options.onUpdate) {
+    store.on("before-update", component, () => {
+      const newValue = selector(store!.state, component.props!);
+      options.onUpdate(newValue);
+    });
   }
   store.updateFunctions[componentId].push(function (): boolean {
     return selectCompareUpdate(store!.state, component.props);
@@ -132,6 +140,9 @@ export function useStore(selector, options: SelectorOptions = {}): any {
   const __destroy = component.__destroy;
   component.__destroy = (parent) => {
     delete store.updateFunctions[componentId];
+    if (options.onUpdate) {
+      store.off("before-update", component);
+    }
     __destroy.call(component, parent);
   };
 
--- a/tests/store_hooks.test.ts
+++ b/tests/store_hooks.test.ts
@@ -571,12 +571,12 @@ describe("connecting a component to store", () => {
     app.state.beerId = 2;
     await nextTick();
     expect(fixture.innerHTML).toBe("<div><span>kwak</span></div>");
-    expect(counter).toBe(1);
+    expect(counter).toBe(0);
 
     store.dispatch("renameBeer", { id: 2, name: "orval" });
     await nextTick();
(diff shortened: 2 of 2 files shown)


ged-odoo:
@seb-odoo

src/store.ts
```diff
@@ -105,13 +110,16 @@ export function useStore(selector, options: SelectorOptions = {}): any {
...
       return true;
     }
     return false;
   }
+  if (options.onUpdate) {
+    store.on('before-update', component,  () => {
+      const newValue = selector(store!.state, component.props!);
+      options.onUpdate(newValue);
+    });
+  }
```
seb-odoo, line 122: For me it's important to call `selector` in this case even if there is no `onUpdate` method.
We use have a custom `isEqual` function that requires that it is called at every render (and especially that is never called if there is no render!) in order to properly save the old value for future compares.

seb-odoo: Note this is also important for `selectCompareUpdate` from OWL, if you don't call it the latest value of scoped `result` won't be saved and the next compare will be wrong. Maybe this handler should write into `result` as well.

seb-odoo, line 122:
```suggestion

    store.on('before-update', component,  () => {
        result = selector(store!.state, component.props!);
        if (options.onUpdate) {
          options.onUpdate(result);
        }
    });
```

seb-odoo: But are we guaranteed that a render will happen when inside `__notifyComponents`? Isn't the goal of notify specifically for the components to do their compare? By calling the selector and assuming there will be a render (from parent?) we might miss a case where `selectCompareUpdate` will actually be called just after, and fail due to updated `result` value?

seb-odoo: As per discussed, the only problem is `onUpdate`, `useStore` will be called before the render so what I said is irrelevant.

aab-odoo (approved):

Issue #825, opened 2021-01-13
Mistake in tutorial reference/observer


qho-odoo:
in section ["Example"](https://github.com/odoo/owl/blob/master/doc/reference/observer.md#example)

```js
const observer = new owl.Observer();
// should be
// const observer = new owl.core.Observer();

observer.notifyCB = () => console.log("update");
const obj = observer.observe({ a: { b: 1 } });

obj.a.b = 2;
```

in section ["Reference"](https://github.com/odoo/owl/blob/master/doc/reference/observer.md#reference)
```js
const observer = new owl.Observer();
const obj = observer.observe({ a: { b: 1 } });

// deepRevNumber() is not defined. Is it changed to revNumber()?
observer.deepRevNumber(obj.a); // 1
obj.a.b = 2;

observer.deepRevNumber(obj.a); // 2
```

aab-odoo:
Closed by https://github.com/odoo/owl/pull/832

Issue #822, opened 2021-01-11
Mistakes in tutorial reference component section "functional components"


qho-odoo:
in https://github.com/odoo/owl/blob/master/doc/reference/component.md#functional-components

```js
// it does not show the name as desired. I guess it should be 
// const Welcome = xml`<h1>Hello, <t t-esc="props.name"/></h1>`;
const Welcome = xml`<h1>Hello, {props.name}</h1>`;

class MyComponent extends Component {
  static template = xml`
        <div>
            <t t-call=${Welcome}/>  // it should be <t t-call="${Welcome}"/> (with quotation marks)
            <div>something</div>
        </div>
    `;
}
```

aab-odoo:
Closed by https://github.com/odoo/owl/pull/832

Issue #821, opened 2021-01-08
Typo in tutorial section "Saving tasks in local storage"


qho-odoo:
In https://github.com/odoo/owl/blob/master/doc/learning/tutorial_todoapp.md#11-saving-tasks-in-local-storage
``` js
function setup() {
  owl.config.mode = "dev";
  const env = {store = makeStore()}; 
  // should be 
  // const env = {store: makeStore()}
  mount(App, { target: document.body, env });
}
```

aab-odoo:
Closed by https://github.com/odoo/owl/pull/832

PR #832, opened 2021-02-03
[DOC] misc small fixes

### LINKED ISSUE #827: Mistake in tutorial reference/slot and unclear error message
in the section ["Example"](https://github.com/odoo/owl/blob/master/doc/reference/slots.md#example)

```xml
<div t-name="SomeComponent">
  <div>some component</div>
<!-- 
should be 
<Dialog title="'Some Dialog'">
with ''
-->
  <Dialog title="Some Dialog">
    <t t-set-slot="content">
      <div>hey</div>
    </t>
    <t t-set-slot="footer">
      <button t-on-click="doSomething">ok</button>
    </t>
  </Dialog>
</div>
```

It will give the error message like `Uncaught (in promise) Error: Invalid generated code while compiling template '__template__6': Unexpected identifier` because  the compiled template tried to find a variable to interpret but it got a string with a space on it, which is an invalid variable name in js.
it should be possible to catch this error in owl and throw something better like invalid prop format

ged-odoo:
closes #827, #825, #822, #821

### DIFF
--- a/doc/learning/tutorial_todoapp.md
+++ b/doc/learning/tutorial_todoapp.md
@@ -663,7 +663,7 @@ function makeStore() {
 
 function setup() {
   owl.config.mode = "dev";
-  const env = {store = makeStore()};
+  const env = {store: makeStore()};
   mount(App, { target: document.body, env });
 }
 ```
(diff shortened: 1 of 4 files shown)


aab-odoo (approved):

seb-odoo:
Just a heads up, it appears the multi-close syntax didn't work, as most of those issues have not been closed.

PR #837, opened 2021-02-04
[REF] component: use a status number instead of flags

ged-odoo:
With this commit, we introduce the status internal number, which
represents the current state of the component. It can currently have 4
values:

- CREATED (component is created and not rendered yet)
- RENDERED (component has been rendered at least once, and is not in the
DOM)
- MOUNTED (component has been rendered, and is in the DOM
- DESTROYED

This status number replaces the isMounted and isDestroyed boolean flags.
It has the advantage of making sure that the component is in a
consistent state (it is no longer possible to be destroyed and mounted,
for example)

Another advantage is that it gives us an easy way to track the fact that
a component has been rendered, but is not in the DOM.  This is a subtle
situation where some various events can happen, and we need to be able
to react to that case.

### DIFF
--- a/src/component/component.ts
+++ b/src/component/component.ts
@@ -44,6 +44,15 @@ interface MountOptions {
   position?: MountPosition;
 }
 
+export const enum STATUS {
+  CREATED,
+  WILLSTARTED, // willstart has been called
+  RENDERED, // first render is completed (so, vnode is now defined)
+  MOUNTED, // is ready, and in DOM. It has a valid el
+  UNMOUNTED, // has a valid el, but is not in DOM
+  DESTROYED,
+}
+
 /**
  * This is mostly an internal detail of implementation. The Meta interface is
  * useful to typecheck and describe the internal keys used by Owl to manage the
@@ -56,8 +65,7 @@ interface Internal<T extends Env> {
   depth: number;
   vnode: VNode | null;
   pvnode: VNode | null;
-  isMounted: boolean;
-  isDestroyed: boolean;
+  status: STATUS;
 
   // parent and children keys are obviously useful to setup the parent-children
   // relationship.
@@ -164,16 +172,18 @@ export class Component<Props extends {} = any, T extends Env = Env> {
         this.env.browser = browser;
       }
       this.env.qweb.on("update", this, () => {
-        if (this.__owl__.isMounted) {
-          this.render(true);
-        }
-        if (this.__owl__.isDestroyed) {
-          // this is unlikely to happen, but if a root widget is destroyed,
-          // we want to remove our subscription.  The usual way to do that
-          // would be to perform some check in the destroy method, but since
-          // it is very performance sensitive, and since this is a rare event,
-          // we simply do it lazily
-          this.env.qweb.off("update", this);
+        switch (this.__owl__.status) {
+          case STATUS.MOUNTED:
+            this.render(true);
+            break;
+          case STATUS.DESTROYED:
+            // this is unlikely to happen, but if a root widget is destroyed,
+            // we want to remove our subscription.  The usual way to do that
+            // would be to perform some check in the destroy method, but since
+            // it is very performance sensitive, and since this is a rare event,
+            // we simply do it lazily
+            this.env.qweb.off("update", this);
+            break;
         }
       });
       depth = 0;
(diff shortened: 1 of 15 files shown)


src/component/component.ts
```diff
@@ -44,6 +44,15 @@ interface MountOptions {
   position?: MountPosition;
 }
 
+export const enum STATUS {
+  CREATED,
+  WILLSTARTED,      // willstart has been called
+  STARTED,          // willStart is completed
+  READY,            // has been patched at least once
```
aab-odoo, line 51: patched? Does it mean that patched has been called?

src/component/component.ts
```diff
@@ -44,6 +44,15 @@ interface MountOptions {
   position?: MountPosition;
 }
 
+export const enum STATUS {
+  CREATED,
```
aab-odoo, line 48: where do you set the status to `CREATED`?

src/component/component.ts
```diff
@@ -317,57 +326,57 @@ export class Component<Props extends {} = any, T extends Env = Env> {
...
-      const currentFiber = __owl__.currentFiber;
-      if (!currentFiber.target && !currentFiber.position) {
-        // this means we have a pending rendering, but it was a render operation,
-        // not a mount operation. We can simply update the fiber with the target
-        // and the position
+      case STATUS.WILLSTARTED:
+      case STATUS.STARTED:
         currentFiber.target = target;
         currentFiber.position = position;
         return scheduler.addFiber(currentFiber);
```
aab-odoo, line 349: not sure about this: are we sure there is a currentFiber? so `started` means there is a pending rendering?

aab-odoo: So `ready` means rendering is done, but not in the DOM?

src/component/component.ts
```diff
@@ -394,11 +403,11 @@ export class Component<Props extends {} = any, T extends Env = Env> {
...
     // currentFiber that is already rendered (isRendered is true), so we are
     // about to be mounted
-    const isMounted = __owl__.isMounted;
+    const status = __owl__.status;
     const fiber = new Fiber(null, this, force, null, null);
     Promise.resolve().then(() => {
-      if (__owl__.isMounted || !isMounted) {
-        if (fiber.isCompleted) {
+      if (__owl__.status === STATUS.MOUNTED || status !== STATUS.MOUNTED) {
+        if (fiber.isCompleted || fiber.isRendered) {
```
aab-odoo, line 410: this is new

src/component/fiber.ts
```diff
@@ -107,6 +107,8 @@ export class Fiber {
    */
   _reuseFiber(oldFiber: Fiber) {
     oldFiber.cancel(); // cancel children fibers
+    oldFiber.target = this.target || oldFiber.target;
+    oldFiber.position = this.position || oldFiber.position;
```
aab-odoo, line 111: new?

src/component/fiber.ts
```diff
@@ -253,7 +255,8 @@ export class Fiber {
           component.__owl__.pvnode!.elm = component.__owl__.vnode!.elm;
         }
       }
-      if (fiber === component.__owl__.currentFiber) {
+      const compOwl = component.__owl__;
```
aab-odoo, line 258: weird

tests/component/un_mounting.test.ts
```diff
@@ -404,6 +404,40 @@ describe("unmounting and remounting", () => {
     expect(fixture.innerHTML).toBe("<div>P2<span>C2</span></div>");
   });
 
+  test("change state while component is mounted in a fragment", async () => {
```
aab-odoo, line 407: maybe split the REF from the FIX?

tests/component/un_mounting.test.ts
```diff
@@ -490,23 +524,24 @@ describe("unmounting and remounting", () => {
...
 
-  test("widget can be mounted on same target, another situation", async () => {
+  test("component can be mounted on same target, another situation", async () => {
     const def = makeDeferred();
     const steps: string[] = [];
 
     class MyWidget extends Component {
       static template = xml`<div>Hey</div>`;
       async willStart() {
+        // console.warn('willstart')
```
aab-odoo, line 544: :-)

src/component/component.ts
```diff
@@ -186,8 +196,7 @@ export class Component<Props extends {} = any, T extends Env = Env> {
       depth: depth,
       vnode: null,
       pvnode: null,
-      isMounted: false,
-      isDestroyed: false,
+      status: 0,
```
aab-odoo, line 199: STATUS.CREATED?

aab-odoo (approved):

PR #847, opened 2021-03-29
Improve tooling ges

### LINKED ISSUE #793: make sure release script rm dist/

### LINKED ISSUE #814: upgrade playground
It's still on an older version, which makes it hard/unreliable to reproduce issues.

Would it be possible to automatically include the upgrade of the playground as par of the release process?

### LINKED ISSUE #829: release script:make sure we are in master (?)
I just made a owl release, but my local code was still on a release branch. Then the release script pushed the release commit on that branch instead of the master... Something needs to be done!

SimonGenin:
Release script errase dist folder
closes #793

Release can only be done on master branch
closes #829

Update the playground owl version on release
closes #814

Add logError method in release to show messages in red.

Check-formatting command added to the release process.

### DIFF
--- a/tools/release.js
+++ b/tools/release.js
@@ -3,9 +3,11 @@ const readline = require("readline");
 const fs = require("fs");
 const exec = require("child_process").exec;
 const chalk = require("chalk");
+const branchName = require('current-git-branch');
 
 const REL_NOTES_FILE = `release-notes.md`;
 const STEPS = 8;
+const branch = "master";
 
 const rl = readline.createInterface({
   input: process.stdin,
@@ -21,6 +23,19 @@ startRelease().then(() => {
 // -----------------------------------------------------------------------------
 
 async function startRelease() {
+  // First check we are on master
+  if (branchName() !== branch) {
+    logError(`You shall not pass! You are not on the ${branch} branch!`)
+    return;
+  }
+  
+  log("Check if code formatting is right...")
+  const checkFormatting = await execCommand("npm run check-formatting");
+  if (checkFormatting !== 0) {
+    logError("Prettier format validation failed. Aborting.");
+    return;
+  }
+
   log(`*** Owl release script ***`);
   log(`Current Version: ${package.version}`);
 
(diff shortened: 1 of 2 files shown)


tools/release.js
```diff
@@ -79,6 +79,11 @@ async function startRelease() {
   log(`Step 5/${STEPS}: building owl...`);
   await execCommand("rm -rf dist/");
   await execCommand("npm run prettier");
+  const checkFormatting = await execCommand("npm run check-formatting");
```
ged-odoo: hmm, it's weird to have a `npm run prettier` just above??

tools/release.js
```diff
@@ -23,11 +24,15 @@ startRelease().then(() => {
 
 async function startRelease() {
   // First check we are on master
-  if (branchName() !== 'master') {
-    logError("you shall not pass! You are not on the master branch!")
+  if (branchName() !== branch) {
+    logError("you shall not pass! You are not on the master " + branch + "!")
```
ged-odoo, line 28: this is wrong I think

tools/release.js
```diff
@@ -23,11 +24,15 @@ startRelease().then(() => {
...
     return;
   }
 
+  // Since we need to jump between branches, ask the use to make sure branches are clean. 
+  // Maybe we could work with the stash, but we keep it simple for now. It seems logic to have a clean branch before a release.
+  await ask(`Make sure all your changes are commited or the script will fail to update the playground. (type anything to continue)`);
```
ged-odoo, line 34: committed

tools/release.js
```diff
@@ -62,26 +74,34 @@ async function startRelease() {
 
   // ---------------------------------------------------------------------------
   log(`Step 4/${STEPS}: creating git commit...`);
+  // For the commit, we make sure prettier has been run.
```
ged-odoo, line 77: i think the prettier check should be done as a separate step (probably one of the first)

tools/release.js
SimonGenin: I removed this. Because it should always work as we commit -am later. After the commit everything should be cleaned before the checkout.

ged-odoo (approved):

PR #852, opened 2021-04-21
[FIX] extensions: guarantee to call transitionend callback

caburj:
There is a possibility that the transitionend event of an element/component
with t-transition directive won't trigger. Though this situation is
difficult to assert, it was observe in odoo runbot for the pos ui.

When the transitionend event is not fired, the callback that removes
the element from the dom won't be called, resulting to a corrupted view.
An example of which is the following:

```html
<div t-if="show" t-transition="fade">Hello</div>
```

If `show` is set to false by some ui action and by any chance the
transitionend event is not fired (perhaps because the transition didn't
actually start or because of completely unknown reason), the div element
will remain in the view -- and this is not desirable.

This commit patches this situation such that if after 50ms that the
transitionend event is supposed to be fired but the event isn't fired, we
force the callback using a setTimeout. This guarantees the call of the
callback that is suppose to remove the element from the view.

### DIFF
--- a/doc/reference/config.md
+++ b/doc/reference/config.md
@@ -6,6 +6,7 @@ global `config` object. It provides two settings:
 
 - [`mode`](#mode) (default value: `prod`),
 - [`enableTransitions`](#enabletransitions) (default value: `true`).
+- [`guaranteeTransitionEnd`](#guaranteetransitionend) (default value: `false`).
 
 ## Mode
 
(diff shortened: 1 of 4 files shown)


ged-odoo:
I only took the commit with the setTimeout and merged it in PR https://github.com/odoo/owl/pull/891 
Thank you

PR #868, opened 2021-06-04
[IMP] qweb: add support for t-tag directive

### LINKED ISSUE #721: Suggestion: new directive t-tag
This directive could be useful when we have a conditionnal tag

e.g.
```xml
<ol t-if="isOrdered">
  I do something
</ol>
<ul t-else="">
  I do the exact same thing as above
</ul>
```

With t-tag we could do
```xml
<t t-tag="isOrdered ? 'ol' : 'ul'">...</t>
```

_The exemple is a real case I encountered_

ged-odoo:
Very useful in some cases, when one needs to define a generic component.

closes #721

### DIFF
--- a/doc/reference/qweb_templating_language.md
+++ b/doc/reference/qweb_templating_language.md
@@ -13,6 +13,7 @@
   - [Setting Variables](#setting-variables)
   - [Conditionals](#conditionals)
   - [Dynamic Attributes](#dynamic-attributes)
+  - [Dynamic Tag Names](#dynamic-tag-names)
   - [Loops](#loops)
   - [Rendering Sub Templates](#rendering-sub-templates)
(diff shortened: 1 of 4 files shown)


doc/reference/qweb_templating_language.md
```diff
@@ -324,6 +326,22 @@ values) or a pair `[key, value]`. For example:
 <div t-att="['a', 'b']"/> <!-- <div a="b"></div> -->
 ```
 
+### Dynamic tag names
+
+When writing generic components or templates, the specific concrete tag for an
+HTML element is not known yet. In those situations, the `t-tag` directive is
+useful. It simply evaluate dynamically an expression to use as a tag name. The
```
aab-odoo, line 333: evaluates

aab-odoo (approved):

PR #883, opened 2021-06-30
[IMP] component: add support for callable expression in event handler

seb-odoo:
Hello @ged-odoo here is a proof of concept for what I believe is a non-breaking change that would allow supporting all callable event handlers until we have arrow functions support.

This would resolve the main issue I have with not being able to access the event for now.

Let me know what do you think.

### DIFF
--- a/src/qweb/extensions.ts
+++ b/src/qweb/extensions.ts
@@ -75,6 +75,7 @@ export function makeHandlerCode(
     // we need to capture every variable in it
     putInCache = false;
     code = ctx.captureExpression(value);
+    code = `const res = (() => { return ${code} })(); if (typeof res === 'function') { res(e) }`;
   }
   const modCode = mods.map((mod) => modcodes[mod]).join("");
   let handler = `function (e) {if (context.__owl__.status === ${STATUS.DESTROYED}){return}${modCode}${code}}`;
--- a/tests/component/component.test.ts
+++ b/tests/component/component.test.ts
@@ -149,6 +149,24 @@ describe("basic widget properties", () => {
     expect(fixture.innerHTML).toBe("<div>1<button>Inc</button></div>");
   });
 
+  test("support for callable expression in event handler", async () => {
+    class Counter extends Component {
+      static template = xml`
+      <div><t t-esc="state.value"/><input type="text" t-on-input="obj.onInput"/></div>`;
+      state = useState({ value: "" });
+      obj = { onInput: (ev) => (this.state.value = ev.target.value) };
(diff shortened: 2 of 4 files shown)


src/qweb/extensions.ts
```diff
@@ -77,7 +77,7 @@ export function makeHandlerCode(
     code = ctx.captureExpression(value);
   }
   const modCode = mods.map((mod) => modcodes[mod]).join("");
-  let handler = `function (e) {if (context.__owl__.status === ${STATUS.DESTROYED}){return}${modCode}${code}}`;
+  let handler = `function (e) {if (context.__owl__.status === ${STATUS.DESTROYED}){return}${modCode}const res = (() => { return ${code} })(); if (typeof res === 'function') { res(e) }}`;
```
seb-odoo, line 80: I couldn't figure out how to store the result of `${code}` without wrapping it in a function.

tests/component/component.test.ts
```diff
@@ -149,6 +149,25 @@ describe("basic widget properties", () => {
...
   });
 
+  test("support for callable expression in event handler", async () => {
+    class Counter extends Component {
+      static template = xml`
+      <div><t t-esc="state.counter"/><button t-on-click="obj.onClick">Inc</button></div>`;
+      state = useState({
+        counter: 0,
+      });
+      obj = { onClick: ev => this.state.counter++ }
```
seb-odoo, line 159: This might not be the best test to cover it, as it doesn't make use of `ev` here. If you are ok with the proposition I guess I should do one with input and reading value of target.

seb-odoo (commented):
2 remarks on current diff

ged-odoo:
@seb-odoo seems promising. I think this can be merged in stable owl version

seb-odoo:
I've updated the test to actually use `ev`, and ran prettier.

Is there anything else that should be done to merge it?

I've noticed in the snapshot that the following case could be improved: `const res = (() => { return utils.getComponent(context)['add'](e); })(); if (typeof res === 'function') { res(e) }`

As it is already generating a function and calling it, the check is not needed. This would be on the `isMethodCall` case, do we want to keep the existing handler line in that case?

ged-odoo:
yes, that was the thing I wanted to check if it can be done. Keeping the existing handler would be better. Otherwise, the commit looks fine.

seb-odoo:
I made the change, it's now injecting the new condition only if it's not a component method call that was built just above.

ged-odoo (approved):

PR #848, opened 2021-04-11
[IMP] router: allow query string in paths

dupski:
We're starting to use OWL for modern front-end development, however the router did not allow navigating to paths with a query string in the URL, e.g.

```js
router.navigate({ path: '/tasks/search?query=test&count=20' })
```
did not match the `/tasks/search` path.

Now it does :)

I've added tests. Let me know if you need anything further.

PR #890, opened 2021-07-05
[IMP] router: allow query string in paths

dupski:
We're starting to use OWL for modern front-end development, however the router did not allow navigating to paths with a query string in the URL, e.g.

```js
router.navigate({ path: '/tasks/search?query=test&count=20' })
```
did not match the `/tasks/search` path.

Now it does :)

I've added tests. Let me know if you need anything further.

### DIFF
--- a/src/router/router.ts
+++ b/src/router/router.ts
@@ -125,7 +125,10 @@ export class Router {
     const initialParams = this.currentParams;
     const result = await this.matchAndApplyRules(path);
     if (result.type === "match") {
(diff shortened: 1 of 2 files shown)


dupski:
@ged-odoo - Re-published to squash commits.
Original PR https://github.com/odoo/owl/pull/848
Thanks for sorting my CLA. Hopefully this can be merged now :)

ged-odoo (approved):

ged-odoo:
thank you for your contribution

PR #894, opened 2021-07-06
ci: update node versions

pparidans:
Only use currently supported Node.js LTS versions.

Reference:
https://nodejs.org/en/about/releases/


ged-odoo:
merged in another pr

PR #897, opened 2021-07-07
[FIX] qweb: properly handle lists in inline expressions

ged-odoo:
Before this commit, Owl inline expressions with a list with multiple
elements such as [a,b,c] was transformed into

[scope['a'], b: scope['b'], scope['c']]

instead of

[scope['a'], scope['b'], scope['c']]

This is due to a previous commit adding support for short object
descriptions such as {a,b}.

To fix this means that we have to keep track of the current group type
for the expression, which is done by using a stack.

### DIFF
--- a/src/qweb/expression_parser.ts
+++ b/src/qweb/expression_parser.ts
@@ -257,16 +257,33 @@ export function compileExprToArray(expr: string, scope: { [key: string]: QWebVar
   const tokens = tokenize(expr);
 
   let i = 0;
+  let stack = []; // to track last opening [ or {
 
   while (i < tokens.length) {
     let token = tokens[i];
     let prevToken = tokens[i - 1];
     let nextToken = tokens[i + 1];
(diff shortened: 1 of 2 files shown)


src/qweb/expression_parser.ts
```diff
@@ -257,16 +257,31 @@ export function compileExprToArray(expr: string, scope: { [key: string]: QWebVar
...
     let nextToken = tokens[i + 1];
+    let groupType = stack[stack.length - 1];
+
+    switch (token.type) {
+      case "LEFT_BRACE":
+        stack.push("{");
+        break;
+      case "LEFT_BRACKET":
+        stack.push("[");
+        break;
```
aab-odoo, line 274: Any reason not to use the token type directly?
```suggestion
      case "LEFT_BRACE":
      case "LEFT_BRACKET":
        stack.push(token.type);
        break;
```

aab-odoo (approved):

PR #898, opened 2021-07-08
[FIX] slots: fix bad interaction between t-call and slots

ged-odoo:
The previous changes in the combine method (used to copy all the
variables defined in the current scope for use in a slot) had the effect
of squashing the prototype chain: instead of `Component -> Obj1 ->
Obj2 -> Obj3`, the combined scope had: `Component -> Obj1'`.

This has an unfortunate interaction with the way t-call is implemented,
which uses the fact that we are in a subscope to add a own
__access_mode__. It depends specifially on the prototype chain, and that
the parent scope may have a different value for that property. But with
the way combine was implemented, we lost all that subtlety since
everything is squashed.

In this commit, we reimplement that function in a way to make sure we
keep the prototype chain structure

### DIFF
--- a/src/qweb/qweb.ts
+++ b/src/qweb/qweb.ts
@@ -137,11 +137,25 @@ const UTILS: Utils = {
     }
     return result;
   },
+  /**
+   * This method combines the current context with the variables defined in a
+   * scope for use in a slot.
+   * 
+   * The implementation is kind of tricky because we want to preserve the
+   * prototype chain structure of the cloned result. So we need to traverse the
+   * prototype chain, cloning each level respectively.
+   */
   combine(context, scope) {
-    const clone = Object.create(context);
+    let clone = context;
+    const scopeStack = [];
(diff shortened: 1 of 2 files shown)


src/qweb/qweb.ts
```diff
@@ -137,12 +137,26 @@ const UTILS: Utils = {
     }
     return result;
   },
+  /**
+   * This method combine the current context with the variables defined in a
```
aab-odoo, line 141: combines

tests/component/slots.test.ts
```diff
@@ -1303,4 +1303,73 @@ describe("t-slot directive", () => {
     document.querySelector("button").click();
     await nextTick();
   });
+
+  test("t-slot in recursive templates", async () => {
+    expect.assertions(1);
```
aab-odoo, line 1308: no need

aab-odoo (approved):

PR #916, opened 2021-10-04
[FIX] remove indeterminism in animation tests

SimonGenin:
Animations have quite a few intricate calls to request next animation
frames. On quick devices, the helper method nextFrame made of 2
request of animation frames would not be enough and race condition
problems would rise.

The fix is simply to ask the nextFrame helper to do a third request
of animation frame in the animation tests, making the test deterministic
on a very fast device. Hopefully, this should be enough.

### DIFF
--- a/tests/animations.test.ts
+++ b/tests/animations.test.ts
@@ -420,29 +420,29 @@ describe("animations", () => {
 
     widget.state.flag = true;
 
-    await nextFrame();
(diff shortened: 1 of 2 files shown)


SimonGenin:
Apparently it is no quite ready..

ged-odoo:
:(

PR #926, opened 2021-10-14
Translation function and t-translation attribute

Polymorphe57:


src/qweb/compiler.ts
```diff
@@ -28,8 +29,8 @@ export type TemplateFunction = (blocks: any, utils: any) => Template;
 
 type BlockType = "block" | "text" | "multi" | "list" | "html";
 
-export function compileTemplate(template: string, name?: string): TemplateFunction {
-  const compiler = new QWebCompiler(template, name);
+export function compileTemplate(template: string, name?: string, translateFn?: (s: string) => string): TemplateFunction {
```
Polymorphe57, line 32: don't like that API

ged-odoo: probably better to put the name and the translateFn as option

src/qweb/compiler.ts
```diff
@@ -143,11 +152,13 @@ export class QWebCompiler {
...
   templateName: string;
   template: string;
+  translateFn?: (s: string) => string;
   ast: AST;
   staticCalls: { id: string; template: string }[] = [];
 
-  constructor(template: string, name?: string) {
+  constructor(template: string, name?: string, translateFn?: (s: string) => string) {
     this.template = template;
+    this.translateFn = translateFn;
```
ged-odoo, line 161: would probably be better to have a fallback: `this.translateFn = translateFn || word => word;` so you don't need to check later on if it exists

Laris15:
Hello everyone, please i have an issue in odoo 14.
I have translated a module i created from English to French by exporting the fr.po file and translating each term in it. But i dont understand why when i change my Odoo language to French all my module is still in English. 

Please can anyone help me

PR #927, opened 2021-10-14
[IMP] qweb: use native Node and Element instead of custom Dom types

sdegueldre:


src/qweb/compiler.ts
```diff
@@ -90,6 +93,14 @@ class BlockDescription {
     }
     return expr;
   }
+
+  asXmlString() {
+    // Can't use outerHTML on text/comment nodes
+    // append dom to any element and use innerHTML instead
+    const t = document.createElement('t');
```
ged-odoo, line 100: maybe it would make sense to do something like this

ged-odoo:
very cool

PR #931, opened 2021-10-15
Owl next dam for lpe

kebeclibre:


Whenrow:
Please name your release twowl instead of owl-next

ged-odoo:
@Whenrow hmm, I think wtotwlwotwott would be clearer...

PR #939, opened 2021-10-20
[IMP] qweb: t-props directive

Polymorphe57:
We reimplement the directive t-props and some tests for it.

### DIFF
--- a/src/qweb/compiler.ts
+++ b/src/qweb/compiler.ts
@@ -905,7 +905,16 @@ export class QWebCompiler {
     for (let p in ast.props) {
       props.push(`${p}: ${compileExpr(ast.props[p]) || undefined}`);
     }
-    const propString = `{${props.join(",")}}`;
+    const propStr = `{${props.join(",")}}`;
+
+    let propString = propStr;
+    if (ast.dynamicProps) {
+      if (!props.length) {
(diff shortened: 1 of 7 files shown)


src/qweb/parser.ts
```diff
@@ -657,6 +658,9 @@ function parseComponent(node: Element, ctx: ParsingContext): AST | null {
     node.removeAttribute("t-component");
   }
 
+  const dynamicProps = node.getAttribute("t-props");
+  node.removeAttribute("t-props");
```
ged-odoo, line 662: i think you don't need to removeattribute here

src/qweb/compiler.ts
```diff
@@ -905,7 +905,12 @@ export class QWebCompiler {
     for (let p in ast.props) {
       props.push(`${p}: ${compileExpr(ast.props[p]) || undefined}`);
     }
-    const propString = `{${props.join(",")}}`;
+    const propStr = `{${props.join(",")}}`;
+
+    let propString = propStr;
+    if (ast.dynamicProps) {
+      propString = `Object.assign({}, ${compileExpr(ast.dynamicProps)}, ${propStr})`;
```
ged-odoo, line 912: would be better to not to the object.assign when there no non-dynamic props

PR #953, opened 2021-11-02
[IMP] component/tags: add inline css tag

brboi:
This add an important feature: defining completely standalone owl components, with the template/style and javascript code together.

### DIFF
--- a/src/component/component_node.ts
+++ b/src/component/component_node.ts
@@ -12,6 +12,7 @@ import {
 } from "./fibers";
 import { applyDefaultProps } from "./props_validation";
 import { STATUS } from "./status";
+import { applyStyles } from "./style";
 
 export function component(
   name: string | typeof Component,
@@ -89,6 +90,9 @@ export class ComponentNode<T extends typeof Component = any> implements VNode<Co
     applyDefaultProps(props, C);
     this.component = new C(props, app.env, this) as any;
     this.renderFn = app.getTemplate(C.template).bind(null, this.component, this);
+    if (C.style) {
+      applyStyles(C);
+    }
     this.component.setup();
   }
 
--- a/src/component/style.ts
+++ b/src/component/style.ts
@@ -1,0 +1,88 @@
+import { Component } from "./component";
+
+export const globalStylesheets: { [key: string]: HTMLStyleElement } = {};
+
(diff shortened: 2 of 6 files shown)


brboi:
I just reintroduced existing code, but IMO we could do better because we never cleanup the `<style/>` elements inside the document's head.

src/component/style.ts
```diff
@@ -0,0 +1,87 @@
...
+export function applyStyles(ComponentClass: typeof Component) {
+  const toApply: { [componentName: string]: string } = {};
+  while (ComponentClass && ComponentClass.style) {
+    if (ComponentClass.hasOwnProperty("style")) {
+      toApply[ComponentClass.name] = ComponentClass.style;
+      delete (ComponentClass as any).style;
+    }
+    ComponentClass = Object.getPrototypeOf(ComponentClass);
+  }
+  for (const [componentName, styleId] of Object.entries(toApply).reverse()) {
```
brboi, line 27: reversing is new here. Any thoughts on that @ged-odoo ?

src/component/style.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+function activateSheet(id: string, name: string) {
+  const sheet = globalStylesheets[id];
+  if (!sheet) {
+    throw new Error(
+      `Invalid css stylesheet for component '${name}'. Did you forget to use the 'css' tag helper?`
+    );
+  }
+  sheet.setAttribute("component", name);
```
ged-odoo, line 39: we should probably use the `data-component` attribute instead

src/component/component_node.ts
```diff
@@ -89,6 +90,7 @@ export class ComponentNode<T extends typeof Component = any> implements VNode<Co
     applyDefaultProps(props, C);
     this.component = new C(props, app.env, this) as any;
     this.renderFn = app.getTemplate(C.template).bind(null, this.component, this);
+    applyStyles(C);
```
ged-odoo, line 93: i would still wrap this in a if: `if (C.styles) {...}`

src/component/style.ts
ged-odoo: toapply => list, push in it, then pop in while loop

ged-odoo (approved):

PR #955, opened 2021-11-03
[FIX] components/fiber: Call patch only if target is valid

abz-odoo:
Currently in some cases, adding an attachment via lognote creates a traceback.
Error : shouldPatch is true while `vnode` is not defined, so `patch()` failed
This is a hotfix correcting this problem by calling `patch()` only if `shouldPatch` is true **and** the `vnode` is set.

opw-2645203

### DIFF
--- a/src/component/fiber.ts
+++ b/src/component/fiber.ts
@@ -243,8 +243,9 @@ export class Fiber {
         }
         component.__patch(target!, fiber.vnode!);
       } else {
-        if (fiber.shouldPatch) {
-          component.__patch(component.__owl__.vnode!, fiber.vnode!);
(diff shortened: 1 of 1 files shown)


src/component/fiber.ts
```diff
@@ -243,8 +243,9 @@ export class Fiber {
         }
         component.__patch(target!, fiber.vnode!);
       } else {
-        if (fiber.shouldPatch) {
-          component.__patch(component.__owl__.vnode!, fiber.vnode!);
+        const { vnode } = component.__owl__!;
```
ged-odoo, line 246: `const vnode = component.__owl__.vnode;`

ged-odoo (approved):

PR #957, opened 2021-11-04
[IMP] qweb: compiler: support t-key on node and component without t-f…

kebeclibre:
…oreach

### DIFF
--- a/src/qweb/compiler.ts
+++ b/src/qweb/compiler.ts
@@ -123,6 +123,7 @@ interface Context {
   preventRoot?: boolean;
   isLast?: boolean;
   translate: boolean;
+  tKeyExpr: string | null;
 }
 
 function createContext(parentCtx: Context, params?: Partial<Context>) {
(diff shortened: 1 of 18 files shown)


src/component/component_node.ts
```diff
@@ -18,9 +19,10 @@ export function component(
   name: string | typeof Component,
   props: any,
   key: string,
+  hasToggle: boolean,
```
ged-odoo, line 22: not sure, but why is the toggler not inserted in the compiled template instead?

kebeclibre: That's what I did in the first commit. I did not like having to set/re-set the key variable in the compiled template. Moreover, in the case we were to have `toggler(key, component(.... key .... ))` the two keys would differ (which is not relevant for the respective algorithms, but still).
This is what we should discuss.

src/qweb/compiler.ts
```diff
@@ -796,6 +808,17 @@ export class QWebCompiler {
   }
 
   compileTKey(ast: ASTTKey, ctx: Context) {
+    let tKeyExpr = compileExpr(ast.expr);
+    if (tKeyExpr.includes("ctx[")) {
```
ged-odoo, line 812: i think we should remove this if

ged-odoo (approved):

ged-odoo (approved):

PR #966, opened 2021-11-12
Remove possibly useless atoms

Polymorphe57:


Polymorphe57:
wrong target

src/reactivity.ts
```diff
@@ -12,18 +12,26 @@ const SOURCE = Symbol("source");
 const OBSERVER = Symbol("observer");
 const KEYS = Symbol("keys");
 const ROOT = Symbol("root");
+const SEED = Symbol("Seed");
+const MAYBE_NOT_A_SEED = Symbol("maybeNotASeed");
 
-export function atom(source: any, observer: Observer) {
+export function atom(source: any, observer: Observer, maybeNotASeed?: any) {
```
Polymorphe57, line 18: the idea is to document only the first two parameters

src/reactivity.ts
```diff
@@ -81,15 +97,14 @@ function createAtom(source: Source, observer: Observer): Atom {
             if (!atoms.has(key)) {
               atoms.set(key, new ObserverSet());
             }
-            atoms.get(key)!.add(self);
+            atoms.get(key)!.add(newAtom);
             keys.add(key);
           }
           //
-          return atom(value, observer);
+          return atom(value, observer, MAYBE_NOT_A_SEED);
```
Polymorphe57, line 104: we can access a value that has been observed by another way: a seed can already exist on that value

src/reactivity.ts
Polymorphe57: and that parameter should be not usable from outside the system

ged-odoo (approved):

PR #970, opened 2021-11-13
Owl next misc

ged-odoo:


src/component/component_node.ts
```diff
@@ -124,7 +124,7 @@ export class ComponentNode<T extends typeof Component = any> implements VNode<Co
     try {
       await Promise.all(this.willStart.map((f) => f.call(component)));
     } catch (e) {
-      handleError(this, e);
+      handleError(this, e as Error);
```
sdegueldre, line 127: I don't think that's correct. Either error handlers should handle non-error objects (willStart can throw anything), or we should wrap non-error objects somewhere into real errors. (IMO the error handlers should just not expect to get a real error, let them deal with a string if willStart throws a string)

(this is not a critique of this change, but of the API of error handlers)

src/compiler/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  // parsing
+  const ast = parse(template);
+
+  // some work
+  const hasSafeContext = !template.includes("t-set") && !template.includes("t-call");
+  const name = options.name || (template.length > 250 ? template.slice(0, 250) + "..." : template);
+
+  // code generation
+  const codeGenerator = new CodeGenerator(name, ast, { ...options, hasSafeContext });
+  const code = codeGenerator.generateCode();
```
sdegueldre, line 22: This could probably just be a function call now?

src/index.ts
```diff
@@ -11,6 +11,7 @@ import {
   toggler,
 } from "./blockdom";
 import { mainEventHandler } from "./component/handler";
+import { EventBus, whenReady, loadFile } from "./utils";
```
sdegueldre, line 14:
```
export * as utils from "./utils";
```

tests/components/basics.test.ts
```diff
@@ -820,86 +821,6 @@ describe("basics", () => {
     await nextTick();
     expect(fixture.textContent!.trim()).toBe("2__3");
   });
-
-  test.skip("subcomponents cannot change observable state received from parent", async () => {
```
sdegueldre, line 824: Do we no longer enforce this?

sdegueldre (approved):

PR #973, opened 2021-11-15
t-out lpe owl next

kebeclibre:


src/blockdom/markup.ts
```diff
@@ -0,0 +1,31 @@
+import { html } from "./html";
```
ged-odoo, line 1: this file should not be in blockdom.it should be at the root /

src/index.ts
```diff
@@ -29,6 +30,7 @@ export const blockDom = {
   toggler,
   createBlock,
   html,
+  safeOut,
 };
```
ged-odoo, line 35: this file should certainly export the `markup` function as well

tests/compiler/t_raw.test.ts
```diff
@@ -84,3 +87,105 @@ describe("t-raw", () => {
...
 });
+
+describe("t-out", () => {
+  let fixture: HTMLElement;
+
+  beforeEach(() => {
+    fixture = makeTestFixture();
+  });
+
+  test("t-out escaoed", () => {
```
ged-odoo, line 98: escaoed => escaped

ged-odoo (commented):
need to adapt all tests from `t_set.test` to use `t-out` instead of `t-raw`

src/blockdom/markup.ts
ged-odoo: hmm, actually, the `markup` function should be in `utils` and the `safeout` code should be in `app/template_helpers`

src/blockdom/markup.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+  let safeKey;
+  let node;
+  if (value instanceof Markup) {
+    safeKey = `string_safe`;
+    node = html(value as string);;
+  } else if (typeof value === "string") {
+    safeKey = "string_unsafe";
+    node = text(value);
+  } else if (value.constructor.name === "Block") {
```
ged-odoo, line 24: this is actually false. it could be any kind of block, not just a Block. what about removing the `if` and assuming that it is a block?

src/index.ts
```diff
@@ -11,7 +11,7 @@ import {
   toggler,
 } from "./blockdom";
 import { mainEventHandler } from "./component/handler";
-import { EventBus, whenReady, loadFile } from "./utils";
+import { EventBus, whenReady, loadFile, markup } from "./utils";
```
ged-odoo, line 14: maybe we should just export them, flat, like everything else. what do you think?

src/utils.ts
```diff
@@ -21,3 +21,25 @@ export async function loadFile(url: string): Promise<string> {
...
+/*
+ * This class just transports the fact that a string is safe
+ * to be injected as HTML. Overriding a JS primitive is quite painful though
+ * so we need to redfine toString and valueOf.
+ */
+export class Markup extends String {
+  constructor(value: any) {
+    super(value);
+    const toString = () => String(value);
+    this.valueOf = toString;
```
ged-odoo, line 34: why do you need to do that?

src/app/template_helpers.ts
```diff
@@ -94,6 +96,29 @@ function shallowEqual(l1: any[], l2: any[]): boolean {
   return true;
 }
 
+/*
+ * Safely outputs `value` as a block depending on the nature of `value`
+ */
+export function safeOut(value: any): ReturnType<typeof toggler> {
```
ged-odoo, line 102: maybe should have another name? like `safeOutput`?

ged-odoo (approved):

PR #975, opened 2021-11-16
loading templates

brboi:


src/compiler/index.ts
```diff
@@ -9,13 +9,21 @@ export type TemplateFunction = (blocks: any, utils: any) => Template;
...
   // some work
-  const hasSafeContext = !template.includes("t-set") && !template.includes("t-call");
-  const name = options.name || (template.length > 250 ? template.slice(0, 250) + "..." : template);
+  const hasSafeContext =
+    template instanceof Node
+      ? !(template instanceof Element) || template.querySelector("[t-set], [t-call]") === null
+      : !template.includes("t-set") && !template.includes("t-call");
+  let name = options.name;
+  if (!name) {
+    const templateString =
```
ged-odoo, line 23: maybe we should not do that. if there is no name, just fallback to something like `template_x`

tests/compiler/simple_templates.test.ts
```diff
@@ -140,49 +140,68 @@ describe("simple templates, mostly static", () => {
 });
 
 describe("loading templates", () => {
```
ged-odoo, line 142: we should move this into another file, maybe `template_set.test.ts`

tests/compiler/simple_templates.test.ts
```diff
@@ -140,49 +140,68 @@ describe("simple templates, mostly static", () => {
...
-  test.skip("can initialize qweb with a string", () => {
-    /*    const templates = `<?xml version="1.0" encoding="UTF-8"?>
+  test("can initialize qweb with a string", () => {
+    const templates = `<?xml version="1.0" encoding="UTF-8"?>
       <templates id="template" xml:space="preserve">
         <div t-name="hey">jupiler</div>
       </templates>`;
-    const qweb = new QWeb({ templates });
-    expect(renderToString(qweb, "hey")).toBe("<div>jupiler</div>");*/
+    const set = new TestContext();
```
ged-odoo, line 148: maybe set => templateSet

tests/compiler/simple_templates.test.ts
```diff
@@ -140,49 +140,68 @@ describe("simple templates, mostly static", () => {
...
 
-  test.skip("can load a few templates from a xml string", () => {
-    /*const data = `<?xml version="1.0" encoding="UTF-8"?>
+  test("can initialize qweb with an XMLDocument", () => {
+    const data = `<?xml version="1.0" encoding="UTF-8"?>
+      <templates id="template" xml:space="preserve">
+        <div t-name="hey">jupiler</div>
+      </templates>`;
+    const xml = new DOMParser().parseFromString(data, "text/xml");
+    const set = new TestContext();
```
ged-odoo, line 159: same as above

sdegueldre (approved):

PR #976, opened 2021-11-16
Next reactivity overhaul sad

sdegueldre:


src/reactivity.ts
```diff
@@ -1,277 +1,163 @@
+import { onWillUnmount } from ".";
```
ged-odoo, line 1: probably better to import from lifecycle hooks to prevent circular dependencies

src/reactivity.ts
```diff
@@ -1,277 +1,163 @@
...
-  const newAtom = atom(source, observer);
-  return [newAtom, unregisterObserver];
-}
-
-function atom(source: any, observer: Observer) {
-  return _atom(source, observer, true);
-}
+// Allows to get the target from an observer (used to reobserving the underlying object)
+const TARGET = Symbol("Target");
+// Special key to subscribe to to be notified of key creation/deletion
```
Polymorphe57, line 6: to to

src/reactivity.ts
```diff
@@ -1,277 +1,163 @@
...
-    if (keepSeeds && atom[SEED]) {
-      continue;
-    }
-    observerAtoms.delete(source);
-    const atoms = sourceAtoms.get(source)!;
-    atoms.get(ROOT)!.delete(atom);
-    for (const key of atom[KEYS]) {
-      atoms.get(key)!.delete(atom);
-    }
+function notifyTarget(target: Target, key: ObjectKey) {
```
Polymorphe57, line 39: callCallbacks?

Polymorphe57: don't like the name

src/reactivity.ts
```diff
@@ -1,277 +1,163 @@
...
+  const keyToCallbacks = targetToKeysToCallbacks.get(target);
+  if (!keyToCallbacks) {
+    return;
   }
-  if (!keepSeeds) {
-    observerSourceAtom.delete(observer);
+  const callbacks = [...(keyToCallbacks.get(key) || [])];
```
Polymorphe57, line 44: why not callbacks = keyToCallbacks.get(key)
then return if !callbacks?

src/reactivity.ts
```diff
@@ -1,277 +1,163 @@
...
-    yield* this.callbackSet;
-  }
 }
 
-// Need to optimize this!
-function isLessOrEqual(node1: ComponentNode, node2: ComponentNode) {
-  let current: any = node1;
-  if (current.level <= node2.level) {
-    return false;
+const proxyCache = new WeakMap<Target, Map<callback, Reactive>>();
```
Polymorphe57, line 71: reactiveCache?

src/reactivity.ts
```diff
@@ -1,277 +1,163 @@
...
+export function reactive<T extends Target>(target: T, notify: callback): Reactive<T> {
```
Polymorphe57, line 72: I would prefere callback: Callback

src/reactivity.ts
```diff
@@ -1,277 +1,163 @@
...
-        willAdd = true;
-      }
+export function batched(callback: callback) {
+  let called = false;
+  return async () => {
+    await Promise.resolve();
+    if (!called) {
+      called = true;
+      callback();
+      await Promise.resolve();
```
Polymorphe57, line 144: why wait here?

src/reactivity.ts
```diff
@@ -1,277 +1,163 @@
...
-  }
-  const unregisterObserver = registerObserver(observer);
-  const newAtom = atom(source, observer);
-  return [newAtom, unregisterObserver];
-}
-
-function atom(source: any, observer: Observer) {
-  return _atom(source, observer, true);
-}
+// Allows to get the target from an observer (used to reobserving the underlying object)
```
Polymorphe57, line 4: from a reactive

src/reactivity.ts
sdegueldre: because `clearReactivesToCallbacks` will modify the set in place and mess with the iteration on it just below, so we need to loop on a copy of the entries.

src/reactivity.ts
```diff
@@ -1,277 +1,163 @@
...
-  if (toNotify) {
-    toNotify.union(observers);
-    return;
-  }
-  toNotify = new ObserverSet();
-  toNotify.union(observers);
-  await Promise.resolve();
-  for (const observer of toClean) {
-    unregisterObserverAtoms(observer, true);
+const batchedRenderFunctions = new WeakMap<ComponentNode, () => void>();
```
Polymorphe57, line 150: Callback

src/reactivity.ts
sdegueldre: I've added some comments for both `Promise.resolve`, let me know if that clears things up

ged-odoo (approved):

PR #986, opened 2021-11-19
SLOTS

Polymorphe57:


src/compiler/code_generator.ts
```diff
@@ -989,17 +965,63 @@ export class CodeGenerator {
...
+    if (ast.dynamicProps) {
+      if (!props.length) {
+        propString = `${compileExpr(ast.dynamicProps)}`;
+      } else {
+        propString = `Object.assign({}, ${compileExpr(ast.dynamicProps)}, ${propStr})`;
+      }
+    }
+
+    let propVar: string;
+    if ((slotDef && (ast.dynamicProps || hasSlotsProp)) || this.dev) {
```
ged-odoo, line 1003: why different in dev mode? i don't think it's important enough to make a difference

Polymorphe57: see if (this.dev) { ... } below

ged-odoo: indeed :)

ged-odoo (approved):

PR #992, opened 2021-11-22
[FIX] components: capture context in prop expressions

sdegueldre:


src/compiler/code_generator.ts
```diff
@@ -344,8 +344,23 @@ export class CodeGenerator {
...
+   * because in compiled code, almost all variables are accessed through the ctx
+   * object. In the case of functions, that lookup in the context can be delayed
+   * which can cause issues if the value has changed since the function was
+   * defined.
+   *
+   * @param expr the expression to capture
+   * @param forceCapture whether the expression should capture its scope even if
+   *  it doesn't contain a function. Useful when the expression will be used as
+   *  a function body.
+   * @returns
```
ged-odoo, line 358: string

ged-odoo (approved):

PR #994, opened 2021-11-22
[FIX] props: prop names can contain -

Polymorphe57:


src/compiler/code_generator.ts
```diff
@@ -967,7 +967,7 @@ export class CodeGenerator {
     const props: string[] = [];
     let hasSlotsProp = false;
     for (let p in ast.props) {
-      props.push(`${p}: ${this.captureExpression(ast.props[p]) || undefined}`);
+      props.push(`'${p}': ${this.captureExpression(ast.props[p]) || undefined}`);
```
ged-odoo, line 970: maybe only do it for name with -:
```js
     const propName = p.includes('-') ? `'${p}'` : p;
      props.push(`${propName}: ${this.captureExpression(ast.props[p]) || undefined}`);
```

Polymorphe57: of course

ged-odoo (approved):

PR #1025, opened 2021-12-07
[FIX] component: make arrow-function capture backwards compatible

sdegueldre:
When fixing the absence of capture for arrow functions passed as props,
we unintentionally introduced a breaking change: bare function calls in
the arrow functions used to be called  with the rendering context as
their this value and this was no longer the case.

This commit fixes that by intentionally not capturing the value of
functions that are called withing the arrow function.


aab-odoo:
robodoo r+

aab-odoo (approved):

PR #1031, opened 2021-12-14
[FIX] fiber, lifecycle: trigger a render during the fiber.complete

kebeclibre:
Have a component which does a render in its onWillPatch, onPatched, onMounted hooks.

Before this commit, the result was incorrect: the second rendering was not taken into account.

After this commit, those renderings are correctly applied.

### DIFF
--- a/src/component/fibers.ts
+++ b/src/component/fibers.ts
@@ -118,9 +118,13 @@ export class RootFiber extends Fiber {
   willPatch: Fiber[] = [];
   patched: Fiber[] = [];
   mounted: Fiber[] = [];
+  // A fiber is typically locked when it is completing and the patch has not, or is being applied.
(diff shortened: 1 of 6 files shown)


src/component/fibers.ts
```diff
@@ -118,9 +118,13 @@ export class RootFiber extends Fiber {
   willPatch: Fiber[] = [];
   patched: Fiber[] = [];
   mounted: Fiber[] = [];
+  // A fiber is typically locked when it is completing and the patch has not, or is being applied.
+  // i.e.: render triggered in onWillUnmout or in willPatch will be delayed
```
ged-odoo, line 122: onWillUnmout

ged-odoo (approved):

PR #1040, opened 2021-12-20
Owl next ged: allow specifying that additional props are ok

ged-odoo:


tests/components/props_validation.test.ts
```diff
@@ -723,6 +723,39 @@ describe("props validation", () => {
...
+      static template = xml`<div>hey</div>`;
+    }
+    class Parent extends Component {
+      static template = xml`<Child message="'m'" otherProp="'o'"/>`;
+      static components = { Child };
+    }
+
+    await expect(async () => {
+      await mount(Parent, fixture, { dev: true });
+    }).not.toThrowError();
```
sdegueldre, line 739: I don't think it works like this. You need .resolves, but apparently .resolves doesn't work with async functions, only .rejects

I'll dig around to see how to test this without it being a pain, in the meantime, what you can do is:
```js
await expect(mount(Parent, fixture, { dev: true })).resolves.toBe(expect.anything());
```

(or just await mount(...) really, if that throws the test will fail though I understand that it doesn't really document that we're asserting that it doesn't throw)

src/component/props_validation.ts
```diff
@@ -19,6 +19,14 @@ export function applyDefaultProps(props: { [key: string]: any }, ComponentClass:
 //------------------------------------------------------------------------------
 // Prop validation helper
 //------------------------------------------------------------------------------
+function getPropDescription(staticProps: any) {
+  if (staticProps instanceof Array) {
+    return Object.fromEntries(
```
ged-odoo, line 24: we need to talk about this and ES2019, to be sure I didn't mess up here

sdegueldre (approved):

PR #1044, opened 2021-12-21
[FIX] portal: do not crash in dev mode

ged-odoo:
Before this commit, the props validation would fail in dev mode because
it did not expect a slot prop.


tests/misc/portal.test.ts
```diff
@@ -24,6 +26,19 @@ function addOutsideDiv(fixture: HTMLElement): HTMLElement {
 
 snapshotEverything();
 
+beforeAll(() => {
```
Polymorphe57, line 29: why in beforeAll and not in the only test that use this?

Polymorphe57 (approved):

PR #1045, opened 2021-12-21
[FIX] useEffect: can depend on dom dependencies

ged-odoo:
Because the dependencies are now computed in patched.

### DIFF
--- a/tests/components/hooks.test.ts
+++ b/tests/components/hooks.test.ts
@@ -17,7 +17,7 @@ import {
   useSubEnv,
   xml,
 } from "../../src/index";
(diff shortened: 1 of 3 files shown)


tests/components/hooks.test.ts
```diff
@@ -396,6 +396,35 @@ describe("hooks", () => {
       ]);
     });
 
+    test("effect can depends on stuff in dom", async () => {
```
sdegueldre, line 399: depends => depend

Also not sure what this tests. It should probably test that the dependency argument is null on first effect run (mounted) and has the element on rerun

sdegueldre (approved):

PR #1046, opened 2021-12-22
[TEST] components: make sure t-ref work with t-if/t-else

ged-odoo:


tests/components/refs.test.ts
```diff
@@ -53,4 +53,32 @@ describe("refs", () => {
...
   });
+
+  test("can use 2 refs with same name in a t-if/t-else situation", async () => {
+    class Test extends Component {
+      static template = xml`
+        <t t-if="state.value">
+          <div t-ref="coucou"/>
+        </t>
+        <t t-else="">
+        <span t-ref="coucou"/>
```
Polymorphe57, line 64: indentation

src/app/template_helpers.ts
```diff
@@ -141,6 +141,19 @@ function bind(ctx: any, fn: Function): Function {
   return boundFn;
 }
 
+type RefMap = { [key: string]: HTMLElement | null };
+type RefSetter = (el: HTMLElement | null) => void;
+
+function multiRefSetter(refs: RefMap, name: string): RefSetter {
```
Polymorphe57, line 147: what about t-if="val === -1" + t-if="val === 1" with val in {-1, 0, 1}?

Polymorphe57: can refs[name] be reset to null/undefined?

sdegueldre (approved):

PR #1047, opened 2021-12-22
[IMP] compiler: scope generated ids to their prefix

sdegueldre:
This means that unrelated ids (eg the id of a template, variable or key)
not longer share the same incrementing counter, meaning that you no
longer see a variable named "v2" unless another variable "v1" was
generated previously.

### DIFF
--- a/src/compiler/code_generator.ts
+++ b/src/compiler/code_generator.ts
@@ -44,7 +44,11 @@ const xmlDoc = document.implementation.createDocument(null, null, null);
 
 class BlockDescription {
   static nextBlockId = 1;
-  static nextDataId = 1;
+  static nextDataIds: { [key: string]: number } = {};
+  static generateId(prefix: string) {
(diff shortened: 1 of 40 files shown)


src/compiler/code_generator.ts
```diff
@@ -44,7 +44,11 @@ const xmlDoc = document.implementation.createDocument(null, null, null);
 
 class BlockDescription {
   static nextBlockId = 1;
-  static nextDataId = 1;
+  static nextDataIds: { [key: string]: number } = {};
+  static generateId(prefix: string) {
+    this.nextDataIds[prefix] = (this.nextDataIds[prefix] || 0) + 1;
+    return `${prefix}${this.nextDataIds[prefix]}`;
```
ged-odoo, line 50: old school js would be simply this: `return prefix + this.nextDataIds[prefix];`

ged-odoo (approved):

PR #1049, opened 2021-12-22
Owl next fix ged

ged-odoo:


tests/components/__snapshots__/t_set.test.ts.snap
```diff
@@ -47,6 +47,150 @@ exports[`t-set slot setted value (with t-set) not accessible with t-esc 2`] = `
...
+  function slot1(ctx, node, key = \\"\\") {
+    ctx = Object.create(ctx);
+    ctx[isBoundary] = 1
+    ctx[\`v\`] = new LazyBlock(value1, ctx, node);
+    let b3 = text(\` in slot \`);
+    let b4 = safeOutput(ctx['v']);
+    return multi([b3, b4]);
+  }
+  
+  function value1(ctx, node, key = \\"\\") {
```
Polymorphe57, line 65: the key param should not be there

Polymorphe57: evaluate in LazyBlock does not pass it to value1

ged-odoo: yeah, it's just the way functions are compiled by the generator. it also happens for slots

src/app/template_helpers.ts
```diff
@@ -101,6 +101,25 @@ function shallowEqual(l1: any[], l2: any[]): boolean {
   return true;
 }
 
+class LazyBlock {
```
Polymorphe57, line 104: strange to call it LazyBlock: no ref to block here

Polymorphe57 (approved):

PR #1054, opened 2022-01-10
[FIX] compiler, component: handle change of t-key before patch

kebeclibre:


kebeclibre:
fixed at https://github.com/odoo/owl/pull/1057

PR #1055, opened 2022-01-11
[FIX] portal: properly handle errors

ged-odoo:
Before this commit, Portal overrode the _render function for its
component node, which means it bypassed the error handling mechanism
that was implemented in that method.  It could have been fixed by
duplicating the error handling code as well, but a better solution in my
opinion is to simply override the renderFn function.  This is closer to
the actual intent of the portal implementation: wrap the result of the
rendering in a VPortal vnode.

### DIFF
--- a/src/portal.ts
+++ b/src/portal.ts
@@ -1,7 +1,6 @@
-import type { ComponentNode } from "./component/component_node";
-import { Component } from "./component/component";
 import { xml } from "./app/template_set";
 import { BDom, text, VNode } from "./blockdom";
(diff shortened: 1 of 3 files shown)


src/portal.ts
```diff
@@ -1,7 +1,6 @@
-import type { ComponentNode } from "./component/component_node";
-import { Component } from "./component/component";
```
Polymorphe57, line 2: Portal overrode?

sdegueldre (approved):

PR #1064, opened 2022-01-14
[FIX] app: t_call recursive template is bound to the correct this

kebeclibre:


src/app/template_set.ts
```diff
@@ -96,7 +96,10 @@ export class TemplateSet {
       const templateFn = this._compileTemplate(name, rawTemplate);
       // first add a function to lazily get the template, in case there is a
       // recursive call to the template name
-      this.templates[name] = (context, parent) => this.templates[name](context, parent);
+      const app = this;
```
ged-odoo, line 99: I feel that it would be more natural to capture `this.templates` instead of `this`:
```js
const templates = this.templates;
this.templates[name] = function (context, parent) {
  return templates[name].call(this, context, parent);
};
```

Polymorphe57 (approved):

PR #1066, opened 2022-01-17
[FIX] typescript: 2 typing issues

ged-odoo:
- useState returned a Reactive, which is true, but in practice makes it
difficult to use outside of internal code for owl
- Component class was no longer generic, which prevented application
code to be properly typed

### DIFF
--- a/src/component/component.ts
+++ b/src/component/component.ts
@@ -1,19 +1,18 @@
-import type { Env } from "../app/app";
 import type { ComponentNode } from "./component_node";
 
 // -----------------------------------------------------------------------------
 //  Component Class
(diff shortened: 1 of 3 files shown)


src/reactivity.ts
```diff
@@ -226,7 +226,7 @@ const batchedRenderFunctions = new WeakMap<ComponentNode, Callback>();
  *  relevant changes
  * @see reactive
  */
-export function useState<T extends object>(state: T): Reactive<T> {
+export function useState<T extends object>(state: T): T {
```
sdegueldre, line 229: I'm not convinced by this change

sdegueldre (approved):

PR #1073, opened 2022-01-18
[FIX] components: cascading render after microtaskTick (makeChildFiber)

kebeclibre:


src/component/component_node.ts
```diff
@@ -167,15 +167,15 @@ export class ComponentNode<T extends typeof Component = typeof Component>
     //   a root fiber to a child fiber in the previous microtick, because it was
     //   embedded in a rendering coming from above, so the fiber will be rendered
     //   in the next microtick anyway, so we should not render it again.
-    if (this.fiber && (current || !fiber.parent)) {
+    if (this.fiber && (current || !fiber.parent) && this.fiber === fiber) {
```
ged-odoo, line 170: it looks like you can simply use this:
```js
    if (this.fiber === fiber && (current || !fiber.parent)) {

```

ged-odoo (approved):

PR #1094, opened 2022-01-24
[FIX] blockdom: fix VHtml patching not setting its html correctly

sdegueldre:


tests/components/basics.test.ts
```diff
@@ -1006,4 +1006,36 @@ describe("t-out in components", () => {
...
+      }
+
+      reverse() {
+        const tmp = this.state.a;
+        this.state.a = this.state.b;
+        this.state.b = tmp;
+      }
+    }
+
+    await mount(Test, fixture);
```
ged-odoo, line 1032: you can remove the `setup` method from `Test` and do this instead:

```js
const comp = await mount(Test, fixture);
```

tests/components/basics.test.ts
```diff
@@ -1006,4 +1006,36 @@ describe("t-out in components", () => {
       "<div>&lt;b&gt;one&lt;/b&gt;<b>one</b>&lt;b&gt;two&lt;/b&gt;<b>two</b>&lt;b&gt;tree&lt;/b&gt;<b>tree</b></div>"
     );
   });
+
+  test("can mount a component with no text", async () => {
```
ged-odoo, line 1010: title does not seem to match the test? maybe it could be "can update t-out content"

ged-odoo (approved):
seems good. just need to adjust the test a little bit

ged-odoo (approved):

PR #1103, opened 2022-01-26
[FIX] component: properly capture expression of t-model

brboi:


tests/components/t_model.test.ts
```diff
@@ -246,6 +246,34 @@ describe("t-model directive", () => {
     expect(fixture.innerHTML).toBe("<div><input><span>test</span></div>");
   });
 
+  test.only("with expression having a changing key", async () => {
```
aab-odoo, line 249: :D

ged-odoo (approved):

PR #1104, opened 2022-01-27
[FIX] compiler: does not modify xml doc in place

ged-odoo:


src/compiler/parser.ts
```diff
@@ -180,23 +180,34 @@ export type AST =
...
   if (!ast) {
-    return { type: ASTType.Text, value: "" };
+    // we clone here the xml to prevent modifying it in place
+    ast = _parse(xml.cloneNode(true) as Element);
+    cache.set(xml, ast);
   }
   return ast;
 }
 
+function _parse(xml: Element): AST {
```
sdegueldre, line 199: Why does this take an Element instead of a Node?

sdegueldre (approved):

PR #1129, opened 2022-02-16
prototype for hydration

ged-odoo:


ged-odoo:
note that the code is horrible and incomplete. It looks like the code for mounting is very similar to the hydration code, so there is certainly a way to factorize these 2 methods in a smart way

PR #1134, opened 2022-02-23
PROTOTYPE: batch renders in next animation frame

ged-odoo:


sdegueldre:
Purpose:
the idea is that when multiple renders happen in the same animation frame, only the last one will be shown. By doing this, we may delay the rendering by up to one frame's worth of time (if a single call to render was made right after an animation frame), which is acceptable in the case of asynchronous renderings, and this can save a lot of work when many renders are queued successively (eg, if keeping a key pressed, events can be emitted faster than the monitor's refresh rate. Rendering more than once per frame can slow down when the browser will repaint, resulting in more renders scheduled on the next frame, and even more on the next until the browser slows down to a crawl)

One complication is that we don't want to delay the rendering by *more* than one frame, meaning that whenever possible, the rendering should be synchronous so that it is flushed/patched in the same frame that it was rendered.

Current progress:
the current solution is probably mostly complete. As it stands, it appears that the remaining failing tests are failing because:
 - they don't expect rendering to ever be purely synchronous, this can mess with the order of the logged lifecycle hooks.
 - they expect that waiting a few microticks will be enough for the render to start, but they should wait for an animation frame instead

PR #1142, opened 2022-03-03
[IMP] component, reactivity: add subscription getter in dev mode

sdegueldre:
This allows to see which objects and which keys in those objects a
component is observing, so that issues with missing renders or extra
renders can be more easily diagnosed.


ged-odoo (approved):

seb-odoo:
Hello, that looks really useful for debugging, it would be nice to add some notes about it in the doc.

PR #1144, opened 2022-03-03
wip

ged-odoo:


src/app/app.ts
```diff
@@ -52,8 +52,10 @@ export class App<
       console.info(DEV_MSG());
       hasBeenLogged = true;
     }
-    const descrs = Object.getOwnPropertyDescriptors(config.env || {});
-    this.env = Object.freeze(Object.defineProperties({}, descrs)) as E;
+    const env = config.env || {};
+    const descrs = Object.getOwnPropertyDescriptors(env);
+    const newEnv = Object.create(Object.getPrototypeOf(env));
```
sdegueldre, line 57: Object.create takes a second optional argument which is a descriptors object, so you can just pass the descriptors here and avoid using Object.defineProperties completely

sdegueldre (approved):

PR #1157, opened 2022-03-15
[IMP] reactivity: synchronous batch context

caburj:
Can be useful for implementing computed values.


caburj:
I'm making this proposal because I'm thinking maybe computed values (derivations) can be implemented something like in this gist: https://gist.github.com/caburj/8cca1de093d4f09c1d536a7b68326871

caburj:
- [ ] rename to `transaction`
- [ ] consider nested `transaction` blocks

PR #1160, opened 2022-03-18
[FIX] reactivity: do not crash when reading reactive frozen objects

### LINKED ISSUE #1158: Reactive frozen objects crash on get

sdegueldre:
This crash was caused by the fact that Proxies *must* return the value of the
property on the target when that property is non-writeable and
non-configurable. Since the reactivity system always attempts to proxify the
value from the target, this crashes.

This commit fixes that by not proxifying such values. This however means that
from that point on, we have escaped the reactivity system and will not
subscribe to any changes in that object or its children.

closes #1158

### DIFF
--- a/src/reactivity.ts
+++ b/src/reactivity.ts
@@ -232,6 +232,11 @@ function basicProxyHandler<T extends Target>(callback: Callback): ProxyHandler<T
       if (key === TARGET) {
         return target;
       }
+      // non-writable non-configurable properties cannot be made reactive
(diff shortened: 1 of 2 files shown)


sdegueldre:
While it is technically possible to get around this limitation by returning a Proxy of an empty object and read from the target, the added layer of indirection would make developer experience worse than it already is for little benefit.

aab-odoo (approved):

ged-odoo (approved):

PR #1164, opened 2022-03-21
[FIX] reactivity: fix crash caused by rendering dead children

sdegueldre:
Previously, when causing a mutation on a state object that would cause both a
parent and a child to be rendered, in some cases, the new state will cause the
parent to no longer render the child. In those cases, it's not uncommon for the
part of the state to which the child is subscribed to be invalid/corrupt and
rendering the child with this state would cause it to crash. While in the
general case of asynchronous renderings, the child or state mutations should be
done in such a way that rendering the child with and "invalid" state will not
crash, we can, in owl, prevent the child from rendering completely when the
parent can be rendered synchronously or at least semi-synchronously (in the
same tick).

This commit will delay the rendering of components by as many microticks as
their depth in the component tree, causing renders that have been scheduled in
the same microtick to render in the order of their depth. Additionally, when a
component renders, it will pre-emptively take account of its children and mark
them as dead if they aren't going to be rendered again, so that we do not have
to wait for the patch to know that the child needs to be destroyed and avoid
the subsequent child render (which may crash).

### DIFF
--- a/src/blockdom/block_compiler.ts
+++ b/src/blockdom/block_compiler.ts
@@ -524,6 +524,17 @@ function createBlockClass(template: HTMLElement, ctx: BlockCtx): BlockClass {
       this.el = el;
       this.parentEl = parent;
     }
+
+    *childNodes() {
+      if (this.children) {
+        for (const child of this.children) {
+          if (child) {
+            yield child;
+            yield* child.childNodes();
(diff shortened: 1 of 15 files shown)


sdegueldre:
Made obsolete by #1167

PR #1169, opened 2022-03-29
[FIX] compiler, components: allow empty slot

kebeclibre:
Before this commit, a t-set-slot that has no content was not even compiled, and thus not passed
to the component for which it has was defined

After this commit, we allow a t-set-slot to have no content (because it can have slot props)


tests/components/slots.test.ts
```diff
@@ -74,6 +74,24 @@ describe("slots", () => {
     expect(fixture.innerHTML).toBe("<span>other text</span>");
   });
 
+  test("simple named and empty slot", async () => {
+    class Child extends Component {
+      static template = xml`<span><t t-slot="default" /><t t-slot="myEmptySlot"/></span>`;
```
sdegueldre, line 79: can you add default-content to `myEmptySlot` to check that it gets overwritten?

sdegueldre (approved):

tests/components/slots.test.ts
kebeclibre: done, and nice catch :) 
the resulting code and templates are even better

ged-odoo (approved):

tests/components/slots.test.ts
```diff
@@ -74,6 +74,43 @@ describe("slots", () => {
...
+      }
+    }
+
+    class Parent extends Component {
+      static template = xml`<Child><t t-set-slot="myEmptySlot" myProp="'myProp text'" /></Child>`;
+      static components = { Child };
+    }
+    await mount(Parent, fixture);
+
+    expect(fixture.innerHTML).toBe("<span>default empty</span>");
```
sdegueldre, line 111: That's not the behaviour I would have expected. What I had in mind was:
- slot is defined but is empty => no default content
- slot is not defined => default content

I guess it's not that important in the end as you can pass a slot-prop to tell the component not to render the slot at all.

PR #1176, opened 2022-04-15
[FIX] concurrency: do not render delayed fibers when cancelled

sdegueldre:
Previously, if a fiber was delayed because one of its ancestors was
rendering, and that fiber was not a root fiber, it would be rendered
after all its ancestors had finished rendering even if one of those
ancestor renderings cancelled it.

This commit fixes that by simply checking that a delayed fiber is still
its component node's current fiber before rendering it.

### DIFF
--- a/src/component/fibers.ts
+++ b/src/component/fibers.ts
@@ -42,6 +42,10 @@ export function makeRootFiber(node: ComponentNode): Fiber {
   return fiber;
 }
 
+function throwOnRender() {
+  throw new Error("Attempted to render cancelled fiber");
+}
+
 /**
  * @returns number of not-yet rendered fibers cancelled
  */
@@ -49,6 +53,7 @@ function cancelFibers(fibers: Fiber[]): number {
   let result = 0;
   for (let fiber of fibers) {
     let node = fiber.node;
+    fiber.render = throwOnRender;
     if (node.status === STATUS.NEW) {
       node.destroy();
     }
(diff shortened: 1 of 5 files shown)


tests/components/concurrency.test.ts
```diff
@@ -3696,6 +3696,42 @@ test("another scenario with delayed rendering", async () => {
   ]).toBeLogged();
 });
 
+test.only("delayed fiber does not get rendered if it was cancelled", async () => {
```
aab-odoo, line 3699: :-)

sdegueldre: (starting to think we really need something to check this in CI)

tests/components/concurrency.test.ts
```diff
@@ -3696,6 +3696,42 @@ test("another scenario with delayed rendering", async () => {
   ]).toBeLogged();
 });
 
+test("delayed fiber does not get rendered if it was cancelled", async () => {
+  class D extends Component {
```
ged-odoo, line 3700: add useLogLifeCycle on each component and check what is logged at every step. so this test will be more robust

src/component/fibers.ts
```diff
@@ -49,6 +49,9 @@ function cancelFibers(fibers: Fiber[]): number {
   let result = 0;
   for (let fiber of fibers) {
     let node = fiber.node;
+    fiber.render = () => {
+      throw new Error("Attempted to render cancelled fiber");
```
ged-odoo, line 53: the function is a constant, you can just define it on top of the file

tests/components/concurrency.test.ts
sdegueldre: Cannot do that as it interferes with the timing heavily.

ged-odoo (approved):

PR #1188, opened 2022-05-16
[IMP] component: only useState on props that are already reactive

sdegueldre:
Previously, components would automatically call useState on their props,
so that changes deeply within props would automatically cause the
component to be rendered. This can be useful when passing a piece of
state to children or descendants.

One problem with this is that all props implicitly become reactive, even
if the object that was passed as a props was not. The problem with that
being that since the original object is not reactive, any change made by
the parent will not go through the reactivity system and the children
won't be notified of the change, in essence, this reactive object is
essentially useless, while having a real cost: traversing reactive
objects creates more reactive objects, and those objects are all
proxies. This is expensive for basically no benefit, while also making
it more difficult to debug code that involves those objects.

This commit fixes that by only calling useState on objects that are
already reactive, allowing the usecase described in the first paragraph
without the drawbacks described in the second.

### DIFF
--- a/src/component/component_node.ts
+++ b/src/component/component_node.ts
@@ -6,6 +6,7 @@ import {
   NonReactive,
   Reactive,
   reactive,
+  toRaw,
   TARGET,
 } from "../reactivity";
 import { batched, Callback } from "../utils";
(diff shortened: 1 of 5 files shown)


src/component/component_node.ts
```diff
@@ -125,7 +126,9 @@ export function component<P extends object>(
 
 type LifecycleHook = Function;
 
-export class ComponentNode<P extends object = any, E = any> implements VNode<ComponentNode<P, E>> {
+export class ComponentNode<P extends { [key: string]: any } = any, E = any>
```
ged-odoo, line 129: maybe use type props instead

src/reactivity.ts
```diff
@@ -78,7 +78,7 @@ export function markRaw<T extends Target>(value: T): NonReactive<T> {
  * @returns the underlying value
  */
 export function toRaw<T extends object>(value: Reactive<T>): T {
-  return value[TARGET] || value;
+  return typeof value === "object" ? (value && value[TARGET]) || value : value;
```
ged-odoo, line 81: not a big fan of this

ged-odoo (commented):
maybe  a test where we check that a prop object is  the same as the parent reference?

aab-odoo (approved):

PR #1227, opened 2022-07-08
Install new pacakages

paulmunyao:


caburj:
1. This is just the same to https://github.com/odoo/owl/pull/1228. Can you close one?
2. What are you trying to accomplish here? The title of the PR doesn't match the changes. Furthermore, the proposed change doesn't make sense. Or are you trying to do something else?

PR #1232, opened 2022-07-15
[FIX] lifecyle_hooks: correctly wrap errors in async code

aab-odoo:
Before this commit, wrapping an error occurring in async code
would result in an unhandledpromise exception, because we created
another promise, that would be rejected and that we didn't catch.

This commit also ensures that we keep the type of error when
wrapping it.


tests/components/error_handling.test.ts
```diff
@@ -264,6 +264,30 @@ describe("errors and promises", () => {
     expect(error!.message).toBe("Tokenizer error: could not tokenize `{ 'invalid: 5 }`");
   });
 
+  test("wrapped errors in async code are correctly caught", async () => {
+    class App extends Component {
```
ged-odoo, line 271: in owl 2 code, I tried to name all `App` component `Root` instead, to avoid confusion with the new `App` class

ged-odoo (approved):

PR #1245, opened 2022-08-24
[FIX] blockdom: fix event_catcher traceback

brboi:
When a parent component has an empty child with a t-on
and an event is triggered inside the parent, blockdom
checks if there is an event_catcher to call.
Doing so, an empty child would cause an error.

This commit fixes that.


brboi:
do not merge for now, it is unclear how we need this fix and what's the cases it covers.

It seems the issue affects only empty child components with a root node and t-foreach directive, but we should dig this.

ged-odoo (approved):

PR #1249, opened 2022-09-06
[FIX] events: correctly call handlers in iframes

sdegueldre:
Previously, event handlers would not work when an app was mounted in an
iframe, this is caused by a guard in the event handler that checks that
the target element is still in the document, but it doesn't check
against the correct document in the case of an iframe.

This commit changes the check to check against the target's
ownerDocument.

### DIFF
--- a/tests/components/event_handling.test.ts
+++ b/tests/components/event_handling.test.ts
@@ -171,4 +171,22 @@ describe("event handling", () => {
     // input is removed when component is destroyed => nothing should happen
     expect([]).toBeLogged();
   });
+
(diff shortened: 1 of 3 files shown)


tests/components/event_handling.test.ts
```diff
@@ -171,4 +171,22 @@ describe("event handling", () => {
     // input is removed when component is destroyed => nothing should happen
     expect([]).toBeLogged();
   });
+
+  test.only("handler works when app is mounted in an iframe", async () => {
```
ged-odoo, line 175: .only

ged-odoo (approved):

PR #1278, opened 2022-10-10
[FIX] runtime: correctly throw an error for duplicate object keys

sdegueldre:
Currently when checking for duplicate keys, we insert the value of the key as is in a set then check for unicity against those. When the key is an object, we check for duplicates based on object identity, whereas the keys are used by owl as strings, and so using objects can cause duplicate key errors that do not throw correctly but crash in the owl internals.

This commit fixes that by making the duplicate checking code serialize the key to string before insertion and when comparing against existing keys.


brboi (approved):
LGTM

PR #1285, opened 2022-10-28
[IMP] doc: improve reactivity documentation

sdegueldre:


doc/reference/reactivity.md
```diff
@@ -2,58 +2,101 @@
...
+the value changes, Owl will update the component. Note that there is nothing special about the
+`state` property, you can name your state variables whatever you want, and you can have multiple of
+them on the same component if it makes sense to do so. This also allows `useState` to be used in custom
+hooks that may require state that is specific to that hook.
 
-  increment() {
-    this.state.value++;
+### Reactive props
+
+Since version 2.0, Owl renders are no longer "deep" by default: a component is only rerenderd by its
```
ged-odoo, line 54: rerendered

doc/reference/reactivity.md
```diff
@@ -2,58 +2,101 @@
...
+parent if its props have changed (using a simple equality test). What if the contents of a props have
```
ged-odoo, line 55: seems weird to have "content" in plural form. not sure it is wrong, but is seems incorrect

doc/reference/reactivity.md
```diff
@@ -64,68 +107,237 @@ The function will be called whenever any tracked value is updated.
...
+console.log(obj1.a); // logs 1, and reads the 'a' key => it is now tracked by observer 1
+console.log(obj2.b); // logs 2, and 'b' is now tracked by observer 2
+obj2.a = 3; // only logs 'observer1', because observer2 does not track a
+obj2.b = 3; // only logs 'observer2', because observer1 does not track b
+console.log(obj2.a, obj1.b); // logs 3 and 3, while the object is observed independently, it is still a single object
 ```
 
-Obviously, one can use `reactive` on the result of a `useState` if wanted, this
-is the proper way to watch for some state changes.
+Because `useState` returns a normal reactive object, it is possible to call `reactive` on the result of a `useState` to observe changes to that object while outside the context of a component, or to call `useState` on reactive objects created outside of components. In those cases, one needs to be careful with regards to the lifetime of those reactive objects, as holding references to these objects may prevent garbage collection of the component and its data even if Owl has destroyed it.
```
ged-odoo, line 129: not really a hard requirement, but i usually try to format the text at roughly 80/100, and not in just one line like this. this makes it easier to read in many cases

doc/reference/reactivity.md
```diff
@@ -64,68 +107,237 @@ The function will be called whenever any tracked value is updated.
...
+this is because in this case, all we care about is that adding or removing notifications in the
+`addNotification` function goes through the reactivity system. The `NotificationContainer` component
+reobserves this object with `useState`, and will be updated whenever notifications are added or removed.
+
+### Store
 
-Also, normal (non-reactive objects) will be directly returned by `toRaw`:
+Centralizing application state is a pretty common want/need in web applications. Because of the way
+the reactivity system works, you can treat any reactive object as a store, and if you call `useState`
+on it, components will automatically observe only the part of the store that they're interested in:
```
ged-odoo, line 282: nice example. also notice that it works with any class, and for now, I think that it's a good pattern:
```js
class TaskList {
  constructor() {
    ...
  }

 addTask() {...}
  toggleTask(){ ...}
}

// later, in a service in odoo
start() {
  const store = reactive(new TaskList());
  return store;
}

// and a utility hook:
function useTaskList() {
  return useState(useService("task"));
}

// and voila, components have a reactive task list that works as a store, but they can call methods on it
```
and

ged-odoo (approved):
nice. i just made some small comments

PR #1297, opened 2022-11-29
[FIX] reactivity: fix issues with reactive objects in proto chain

sdegueldre:
Previously, when there was a reactive object in the prototype chain of a different object, it would get notified of the first write to a property that existed on the reactive object but did not yet exist on the other object, despite the value of that property not actually getting written to the reactive and hence the value not getting changed.

This was caused by the fact that we assumed that Reflect.set would set the value on the target, when in fact, the value is set on the object that underlies the `receiver`. When a reactive object is part of the prototype chain, the first write on a key triggers the set trap but the receiver is not the reactive object, and so Reflect.set doesn't modify the target, but adds the new key to the object that's lower in the prototype chain.

This commit fixes that by actually reading the value from the target instead of assuming that it was changed by Reflect.set

This commit also removes the special symbols SKIP and TARGET, as there is no reliable way to check whether these keys are present on the object itself or on its prototype chain, which can cause issue when trying to create reactive objects from objects with other reactive objects in their prototype chain, or to create reactive objects from objects with non-reactive objects in their prototype chain. To solve this problem, we simply use a WeakMap that maps reactives to their targets, and a WeakSet that contains all skipped objects.

### DIFF
--- a/tests/reactivity.test.ts
+++ b/tests/reactivity.test.ts
@@ -9,7 +9,7 @@ import {
   markRaw,
   toRaw,
 } from "../src";
-import { reactive, Reactive } from "../src/runtime/reactivity";
+import { reactive } from "../src/runtime/reactivity";
 import { batched } from "../src/runtime/utils";
 import {
   makeDeferred,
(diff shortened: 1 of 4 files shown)


tests/reactivity.test.ts
```diff
@@ -338,6 +338,7 @@ describe("Reactivity", () => {
     expect(n).toBe(1);
     state2.a;
     state.a = 2;
+    debugger;
```
ged-odoo, line 341: debugger

ged-odoo (approved):

PR #1299, opened 2022-11-30
[FIX] runtime/fibers: render a parent when self is in error

### LINKED ISSUE #1298: Error handling in components: re-render a parent with onError
usecase as test:
```xml

  test.only("re-render a parent with onError", async () => {
    class Boom extends Component {
      static template = xml`<div />`
      setup() {
        onWillStart(() => {
          throw new Error("Boom Error")
        })
      }
    }

    const steps: string[] = [];
    class BoomWrapper extends Component {
      static template = xml`<Boom t-if="!lol"/>`
      static components = {Boom}
      setup() {
        onError(() => {
          console.debug("PRou")
          steps.push("onError in child");
          this.props.onError()
        })
      }
    }

    class Classic extends Component {
      static template = xml`<BoomWrapper t-if="!inError" onError="onErrorAsProps"/><div t-else="" />`
      static components = { BoomWrapper }
      inError: Boolean = false;
      setup() {
        onMounted(() => {
          steps.push("mounted")
        })
      }
      onErrorAsProps() {
        this.inError = true;
        this.render(true);
      }
    }

    class App extends Component {
      static template = xml`<Classic />`
      static components = { Classic }
    }

    await mount(App, fixture);
    expect(steps).toBe(["onError in child", "mounted"]);
  })

```

kebeclibre:
Have a Component A which instantiate another one (B) which also have some Children (C). Have B be the one to handle the errors from the Children via the onError hook. This hook should call a props coming from A. The real error handler is on A then, passed to B via props.

This handler naturally sets a flag on A, and triggers a render.

Before this commit, the mounting of A never resolved as the fiber that was recycled in A was still part of a root one which was flagged as being in error.

After this commit, the A component mounts correctly.

Closes #1298

### DIFF
--- a/tests/components/error_handling.test.ts
+++ b/tests/components/error_handling.test.ts
@@ -9,6 +9,7 @@ import {
   onRendered,
   onWillUnmount,
   useState,
+  reactive,
   xml,
 } from "../../src/index";
 import {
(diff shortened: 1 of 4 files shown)


tests/components/error_handling.test.ts
```diff
@@ -785,6 +786,131 @@ describe("can catch errors", () => {
...
+      }
+    }
+
+    const steps: string[] = [];
+    class BoomWrapper extends Component {
+      static template = xml`<Boom t-if="state.errorTree === 'error'" /><t t-else="" t-esc="state.errorTree" />`;
+      static components = { Boom };
+      state: any;
+      setup() {
+        this.state = useState(this.props.state);
```
ged-odoo, line 851: you don't need to do that in theory, it's done internally by owl

ged-odoo (approved):

ged-odoo:
looks good to me, i just made a small comment. you can take it into account or not, then merge this

sdegueldre:
This is still in limbo @kebeclibre

PR #1302, opened 2022-12-02
[FIX] compiler: correctly escape backslashes when emitting block string

### LINKED ISSUE #1300: Attribute values treat \ as escape
Template attribute values use `\` character as escape. I don't know if this is expected behavior or not? I can imagine changing this could be a compatibility breaking change. It is also possible to work around by simply escaping the escape character. I just wanted to put it out there because it lead to unexpected behavior when I was creating a template.

Template:
```xml
<input type="text" pattern="^\S+$">
```

Output:
```html
<input type="text" pattern="^S+$">
```

sdegueldre:
Previously, when a template contained backslashes, they were not escaped when creating the blockstring, meaning they would be interpreted as an escape sequence within the JS string. This means that backslashes preceding most characters were completely ignored and didn't end up in the final block, double backslashes were collapsed to a single one, and backslashes that constituted valid escape sequences in JS would be treated as those (eg, \n would be a newline).

When creating a JS string expression from a string value, all characters with special meaning in JS should be escaped, we were correctly escaping backticks as these would unexpectedly close the string if not escaped, but forgot to escape backslashes. This commit fixes that.

closes #1300

### DIFF
--- a/tests/compiler/__snapshots__/attributes.test.ts.snap
+++ b/tests/compiler/__snapshots__/attributes.test.ts.snap
@@ -430,6 +430,19 @@ exports[`attributes static attributes on void elements 1`] = `
 }"
 `;
 
+exports[`attributes static attributes with backslash or backtick 1`] = `
+"function anonymous(app, bdom, helpers
+) {
+  let { text, createBlock, list, multi, html, toggler, comment } = bdom;
+  
+  let block1 = createBlock(\`<div foo=\\"\\\\\\\\a\\" bar=\\"\\\\\\\\n\\" baz=\\"\\\\\`\\"/>\`);
+  
+  return function template(ctx, node, key = \\"\\") {
+    return block1();
+  }
+}"
+`;
+
 exports[`attributes static attributes with backticks 1`] = `
 "function anonymous(app, bdom, helpers
 ) {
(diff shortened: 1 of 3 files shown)


ged-odoo (approved):

tests/compiler/__snapshots__/attributes.test.ts.snap
```diff
@@ -430,6 +430,19 @@ exports[`attributes static attributes on void elements 1`] = `
 }"
 `;
 
+exports[`attributes static attributes with backslash or backtick 1`] = `
+"function anonymous(app, bdom, helpers
+) {
+  let { text, createBlock, list, multi, html, toggler, comment } = bdom;
+  
+  let block1 = createBlock(\`<div foo=\\"\\\\\\\\a\\" bar=\\"\\\\\\\\n\\" baz=\\"\\\\\`\\"/>\`);
```
ged-odoo, line 438: lol at this

sdegueldre: JS string that contains a JS function that contains a JS string => backslashes are escaped 3 times => 2^3 backslashes per backslash. Very cool.

PR #1314, opened 2023-01-09
[IMP] slots: make defaultContent of slot available to slot as `zero`

sdegueldre:
In some cases, it is useful for a slot to be able to keep the default content of the slot it is supposed to replace and simply add extra content instead of simply replacing it entirely. This commit makes the default content of a slot available in the context as `zero` similar to how t-call can access the content that was inside the t-call.


ged-odoo (commented):
looks weirdly elegant. I feel unsure that this is what we want, but it makes sense. also, if we do this, the doc should be updated

PR #1339, opened 2023-01-27
[FIX] reactivity: improve performance for long-lived reactives

sdegueldre:
Previously, when having a long lived reactive object and writing a lot of keys to it, clearing the reactive subscriptions on that object would slow down as time went on. This is because when clearing a target's subsctiptions for a callback, we look at all the keys that are observed, and for all the observed keys, we remove the callback from the set of callbacks observing that key. The problem arises from the fact that after doing that, even if the set of callbacks observing that key is now empty, we don't remove the key from the observed keys, meaning that any further clearing of subscriptions will have to iterate over that key to attempt to clear the callbacks even if there are none.

This commit fixes this problem by simply removing the key from the observedKeys if there are no longer any callbacks observing it.

This commit also improves performance for bare reactives (reactives created with no callback). The initial implementation simply creates a default empty callback when creating a reactive, and treats it like any other, meaning it can observe keys and be notified of changes even though we know in advance that it does nothing. This can compound with the previous issue when you're doing a lot of manipulations on a bare reactive internally for the sole purpose of notifying outside observers, as this creates a lot of useless work in the reactivity system.

### DIFF
--- a/src/runtime/reactivity.ts
+++ b/src/runtime/reactivity.ts
@@ -1,8 +1,13 @@
-import { Callback } from "./utils";
+import type { Callback } from "./utils";
 import { OwlError } from "./error_handling";
 
 // Special key to subscribe to, to be notified of key creation/deletion
 const KEYCHANGES = Symbol("Key changes");
+// Used to specify the absence of a callback, can be used as WeakMap key but
+// should only be used as a sentinel value and never called.
+const NO_CALLBACK = () => {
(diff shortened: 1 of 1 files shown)


sdegueldre:
Worth noting that for best performance, bare reactives could use a different proxy handler that never even tries to observe the key, saving a function call in a few places, but this special case would make the code even harder to understand than it already is, especially since we would need separate handlers for regular objects and collections.

ged-odoo (approved):

PR #1341, opened 2023-01-27
[FIX] hooks: fix useRef in Firefox 109+

Dj0ulo:
[FIX] hooks: fix useRef in Firefox 109+

Since Firefox 109, appending an HTMLElement into an iframe changes the
prototype of that element from the HTMLElement of the window where it
was created to the HTMLElement of the window of the iframe into which it
is appended. This causes the `instanceof HTMLElement` check in `useRef`
to fail, as `HTMLElement` refers to the window in which owl is declared.

This commit fixes that by adding an extra `instanceof` check that uses
the element's ownerDocument window.


sdegueldre:
for reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1813499

PR #1366, opened 2023-03-05
[IMP] .gitignore: ignore tools/devtools build folder

ged-odoo:
Otherwise, it is annoying to have build files that should never be committed anyway

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ node_modules
 
 # Extras temp file
 /tools/owl.js
+/tools/devtools/build
 
 release-notes.md
 


sdegueldre:
I just realized that apparently this was not added by the PR that added the devtools and is missing, never realized since I added the folder to my git exclude temporarily. We probably want part of this PR to be merged.

ged-odoo:
hmm, not sure what you mean. Since then, the devtools build code has been changed, and the devtools are now created in dist/ which are git-ignored. If you still have a `/tools/devtools/build`, you can probably safely delete it

PR #1391, opened 2023-03-30
[FIX] devtools: Fix first release's small bugs

juliusc2066:
This PR address multiple small bugs that were detected in the early first release of the extension

The first commit adds box-sizing: border-box to the highlight elements created by the devtools on the page. This was making some of the boxes overflow on some pages (like on POS sessions) because the border-box box-sizing was not the default behavior on these pages.

The second commit hides the collapse all button when the profiler is in Events log mode since it doesn't do anything in this case.

The third commit increases the vertical padding of the search bar in order for it to be easier to click on.

### DIFF
--- a/tools/devtools/src/devtools_app/devtools_window/profiler_tab/profiler_tab.xml
+++ b/tools/devtools/src/devtools_app/devtools_window/profiler_tab/profiler_tab.xml
@@ -10,7 +10,7 @@
           <option t-att-selected="store.eventsTreeView" value="Tree">Tree view</option>
           <option t-att-selected="!store.eventsTreeView" value="List">Events log</option>
         </select>
-        <i title="Collapse All" class="fa fa-list me-1" t-on-click="() => this.store.collapseAll()"></i>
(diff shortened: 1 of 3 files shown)


tools/devtools/src/devtools_app/devtools_window/profiler_tab/profiler_tab.xml
```diff
@@ -10,7 +10,7 @@
           <option t-att-selected="store.eventsTreeView" value="Tree">Tree view</option>
           <option t-att-selected="!store.eventsTreeView" value="List">Events log</option>
         </select>
-        <i title="Collapse All" class="fa fa-list me-1" t-on-click="() => this.store.collapseAll()"></i>
+        <i title="Collapse All" type="button" class="fa fa-list me-2" t-on-click="() => this.store.collapseAll()" t-attf-style="{{store.eventsTreeView ? '' : 'visibility: hidden;'}}"></i>
```
ged-odoo, line 13: why didn't you just do a `t-if`? i guess it is intended to keep the empty space to prevent stuff from moving when switching mode?

juliusc2066: indeed

ged-odoo (approved):

PR #1395, opened 2023-04-05
[IMP] runtime: allow validating object values using a type description

sdegueldre:
In JS, objects are often used as a mapping that gives fast lookup on the keys, in those cases, the keys themselves are not known in advance, but the values may have a shape that they're expected to follow. This commit adds support for a `values` key in schema for objects, which will be checked against all values in the object during validation.

### DIFF
--- a/doc/reference/props.md
+++ b/doc/reference/props.md
@@ -212,6 +212,7 @@ For each key, a `prop` definition is either a boolean, a constructor, a list of 
   - `type`: the main type of the prop being validated
   - `element`: if the type was `Array`, then the `element` key describes the type of each element in the array. If it is not set, then we only validate the array, not its elements,
   - `shape`: if the type was `Object`, then the `shape` key describes the interface of the object. If it is not set, then we only validate the object, not its elements,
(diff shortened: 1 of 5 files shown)


doc/reference/props.md
```diff
@@ -212,6 +212,7 @@ For each key, a `prop` definition is either a boolean, a constructor, a list of
   - `type`: the main type of the prop being validated
   - `element`: if the type was `Array`, then the `element` key describes the type of each element in the array. If it is not set, then we only validate the array, not its elements,
   - `shape`: if the type was `Object`, then the `shape` key describes the interface of the object. If it is not set, then we only validate the object, not its elements,
+  - `values`: if the type was `Object`, then the `values` key describes the interface of values in the object, this allows validating objects that are used as mappings, where keys are not known in advance but the shape of the values is.
```
ged-odoo, line 215: could you add also an example somewhere below? maybe around line 290

ged-odoo (approved):
very nice. i just left a small comment

doc/reference/props.md
sdegueldre: Done!

ged-odoo (approved):

PR #1396, opened 2023-04-06
[IMP] Allow app to be mounted in shadow DOM

tsm-odoo:
This commit allow owl Apps to be mounted into a shadow DOM.

### DIFF
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -84,7 +84,10 @@ export class App<
     this.props = config.props || ({} as P);
   }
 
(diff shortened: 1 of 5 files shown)


ged-odoo (commented):
it would help to have a simple test (maybe in a new test suite shadow_dom). also, if you don't mind, could you add a second test with a ref in it, because I suspect refs are broken

tsm-odoo:
@ged-odoo, you were right about ref,s I updated the PR to make it work. I also added tests covering:
- the base case
- event handling
- use ref usage

PR #1398, opened 2023-04-07
[IMP] devtools: Add devtools documentation

juliusc2066:
This commit adds the documentation for the devtools extension.


ged-odoo:
@juliusc2066 it's red! test for doc links fail, probably because it tries to read content of images

ged-odoo (approved):

PR #1406, opened 2023-04-14
[REF] devtools: Better messages forwarding

juliusc2066:
This commit improves how messages are relayed from the page to the devtools by using a specific port for each instance of the owl devtools panel and refreshing it through the heartbeat message. This allows to prevent sending messages to every devtools instances and sorting the messages inside.


tools/devtools/src/background.js
```diff
@@ -75,8 +75,23 @@ browserInstance.runtime.onMessage.addListener(async (message, sender, sendRespon
     return true;
   } else if (message.type === "owlStatus") {
     setOwlStatus(message.data);
-    // Dummy message to test if the extension context is still valid
+    // Refresh panel connection timeout
   } else if (message.type === "keepAlive") {
+    const panel = activePanels.get(message.id);
+    if (panel) {
+      clearTimeout(panel.expirationTimeout);
```
ged-odoo, line 82: i feel like the code is not properly designed. we should probably factorize the communication code out of the rest of the code at some point

ged-odoo (approved):

PR #1412, opened 2023-04-24
[IMP/FIX] playground: fix and improve various things

sdegueldre:
The playground hasn't been touched for a while and a few things were
broken before this commit:
- Exporting as a standalone web application was broken because it still
attempted to load the templates like it was done in owl 1
- Resizing the editor tabs still tried to use `this.trigger`

While fixing the export feature, the files needed by the app were
factored out of hardcoded strings and into real files, as this makes it
easier to maintain, since these files get syntax highlighting.

The samples received the same treatment: there is now a `samples` folder
containing all the samples, it also contains a jsconfig, giving
autocompletion on owl functions while editing the files, and allowing
the owl import to look how it would when using owl as a node_module.
In the browser, this import is translated to the relative path of the
owl module using an import map.

The second commit moves the github page files to a docs folder on the master branch to make deploying the playground easier.


tools/release.js
```diff
@@ -131,55 +138,19 @@ async function startRelease() {
...
   }
 
-  log(`Step 8/${STEPS}: publishing module on npm...`);
-  await execCommand("npm run publish");
+  log(`Step ${step++}/${STEPS}: publishing module on npm...`);
+  await execCommand("npm publish");
 
   log("Owl Release process completed! Thank you for your patience");
   await execCommand(`gh release view`);
   await execCommand(`gh release view -w`);
```
ged-odoo, line 153: probably need a step to copy last version of owl.iife.js into `docs/assets/owl.js`?

ged-odoo (approved):
love to see the import map.
i just did a few small comments, nothing important

tools/release.js
sdegueldre: There is one, except it's `owl.es.js` since I moved to es modules instead of using the owl global (which allows us to use import maps)

PR #1417, opened 2023-04-24
[DOC] Fix a code bug in the example of slots

myearn4:
The value of activeTab should be tab_index rather than tab.


ged-odoo (approved):

ged-odoo:
Thank you

PR #1419, opened 2023-04-25
[FIX] devtools: Missing global owl

juliusc2066:
This commit adapts the behavior of the devtools global hook to get the toRaw and reactive functions from the __OWL_DEVTOOLS__ variable instead of the global owl when it is missing. Also adds the functions to the __OWL_DEVTOOLS__ variable for the former to work. This allows the devtools to work in environments where owl is loaded through ES modules.
Also quickly fixes iframes detection update on page reload.

### DIFF
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -7,6 +7,7 @@ import { Scheduler } from "./scheduler";
 import { validateProps } from "./template_helpers";
 import { TemplateSet, TemplateSetConfig } from "./template_set";
 import { validateTarget } from "./utils";
+import { toRaw, reactive } from "./reactivity";
 
 // reimplement dev mode stuff see last change in 0f7a8289a6fb8387c3c1af41c6664b2a8448758f
 
@@ -39,6 +40,8 @@ declare global {
       apps: Set<App>;
       Fiber: typeof Fiber;
       RootFiber: typeof RootFiber;
+      toRaw: typeof toRaw;
+      reactive: typeof reactive;
     };
   }
 }
@@ -47,6 +50,8 @@ window.__OWL_DEVTOOLS__ ||= {
   apps: new Set<App>(),
   Fiber: Fiber,
   RootFiber: RootFiber,
+  toRaw: toRaw,
+  reactive: reactive,
 };
 
 export class App<
--- a/tools/devtools/src/page_scripts/owl_devtools_global_hook.js
+++ b/tools/devtools/src/page_scripts/owl_devtools_global_hook.js
@@ -11,6 +11,10 @@
       this.Fiber = window.__OWL_DEVTOOLS__.Fiber;
       // Same but for RootFiber
       this.RootFiber = window.__OWL_DEVTOOLS__.RootFiber;
+      // This is for retrocompatibility purposes since new versions of owl should always expose toRaw and reactive
+      // in __OWL_DEVTOOLS__
+      this.toRaw = window.__OWL_DEVTOOLS__.toRaw ?? window.owl?.toRaw;
+      this.reactive = window.__OWL_DEVTOOLS__.reactive ?? window.owl?.reactive;
       // Set to keep track of the fibers that are in the flush queue
       this.queuedFibers = new WeakSet();
       // Set to keep track of the HTML elements we added to the page
@@ -400,7 +404,7 @@
       let targetToKeysToCallbacks;
 
       // Step 1: extract internal values from owl
-      const obj = owl.reactive({}, () => {});
+      const obj = self.reactive({}, () => {});
       let count = 0;
       WeakMap.prototype.get = function () {
         count++;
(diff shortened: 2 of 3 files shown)


tools/devtools/src/page_scripts/owl_devtools_global_hook.js
```diff
@@ -388,7 +388,8 @@
       let targetToKeysToCallbacks;
 
       // Step 1: extract internal values from owl
-      const obj = owl.reactive({}, () => {});
+      const obj =
+        window?.owl?.reactive({}, () => {}) || window.__OWL_DEVTOOLS__.reactive({}, () => {});
```
sdegueldre, line 392: remove the empty callback, it's implicit and there is a performance optimization if you don't specify it. 

I don't think this solution works well with multiple instances of owl (eg one in the main page and one in an iframe) because they each have their own version of reactive that captures different weakmaps, so the toRaw of one owl doesn't work with another owl.

sdegueldre (dismissed):

tools/devtools/src/page_scripts/owl_devtools_global_hook.js
juliusc2066: There is one hook per instance of owl so I don't think it's really problematic: the iframe's __OWL_DEVTOOLS__ window variable will be unrelated to the main frame's one so even if the code is different, there is no interaction between both

juliusc2066: For the empty callbacks, i'm not sure about why it matters since Géry did it but it does indeed matter and won't work without it

src/runtime/app.ts
```diff
@@ -39,6 +40,8 @@ declare global {
       apps: Set<App>;
       Fiber: typeof Fiber;
       RootFiber: typeof RootFiber;
+      toRaw: <T extends object, U extends T>(value: U | T) => T;
+      reactive: <T extends object>(target: T, callback?: Callback) => T;
```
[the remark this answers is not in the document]
juliusc2066: oops

tools/devtools/src/page_scripts/owl_devtools_global_hook.js
sdegueldre: > For the empty callbacks, i'm not sure about why it matters since Géry did it but it does indeed matter and won't work without it
Then it's a bug and we should look into it.

> There is one hook per instance of owl so I don't think it's really problematic: the iframe's OWL_DEVTOOLS window variable will be unrelated to the main frame's one so even if the code is different, there is no interaction between both

Hm, I guess that's true, you would need to have 2 separate instances of owl in the same window which should not be a frequent use case. You should probably extract reactive and toRaw just once at the top level. If you keep the `window.owl` method, it should probably be as a fallback and mention that this is for backwards compatibility with owl <=2.1.2 that doesn't expose these in the devtools hook yet.

ged-odoo: the empty callback is necessary to make owl go through the non optimized path so the horrible hack just below works. So this PR does not worsen the current situation.

Now, the case with multiple owl versions is still a problem (not with an iframe though, since the iframe will load a separate copy of the owldevtools global hook). I strongly suspect that it does not work now, and that it would require some other changes in the way we expose owl internal values. I guess it should be solved in a later issue

tools/devtools/src/page_scripts/owl_devtools_global_hook.js
```diff
@@ -1484,7 +1485,9 @@
         return;
       }
       if (objectType === "subscription") {
-        owl.reactive(obj)[key] = value;
+        window?.owl?.reactive
+          ? owl.reactive(obj)[key]
+          : (window.__OWL_DEVTOOLS__.reactive(obj)[key] = value);
```
sdegueldre, line 1490: these parens are not placed correctly, this problem would go away if you extracted reactive and toRaw once during construction.

tools/devtools/src/page_scripts/owl_devtools_global_hook.js
```diff
@@ -51,6 +51,12 @@
       this.traceSubscriptions = false;
       this.requestedFrame = false;
       this.eventsBatch = [];
+      // This is for retrocompatibility purposes since new versions of owl should always expose toRaw and reactive
+      // in __OWL_DEVTOOLS__
+      this.toRaw = window?.owl?.toRaw ? window.owl.toRaw : window.__OWL_DEVTOOLS__?.toRaw;
+      this.reactive = window?.owl?.reactive
+        ? window.owl.reactive
+        : window.__OWL_DEVTOOLS__?.reactive;
```
sdegueldre, line 59:
```suggestion
      this.toRaw = window.__OWL_DEVTOOLS__.toRaw ?? window.owl?.toRaw;
      this.reactive = window.__OWL_DEVTOOLS__.reactive ?? window.owl?.reactive;
```
`window.__OWL_DEVTOOLS__` is known to exist (it's used a little higher). This code should probably be located with the other code that extracts stuff from `__OWL_DEVTOOLS__`

ged-odoo (approved):

ged-odoo (approved):

PR #1439, opened 2023-05-24
[IMP] devtools: Add blacklist to components toggle

juliusc2066:
This commit adds the functionnality to add components to the toggle blacklist so that it won't be expanded by default when launching or reloading the devtools components tree. Also change chrome devtools version to 1.1 for the webstore.

### DIFF
--- a/tools/devtools/manifest-chrome.json
+++ b/tools/devtools/manifest-chrome.json
@@ -1,6 +1,6 @@
 {
   "name": "Owl devtools",
-  "version": "1.0",
+  "version": "1.1",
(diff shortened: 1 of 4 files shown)


tools/devtools/manifest-chrome.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Owl devtools",
-  "version": "1.0",
```
ged-odoo, line 3: make separate commit

ged-odoo (approved):

PR #1440, opened 2023-05-24
[FIX] devtools: fix context menus flickering

juliusc2066:
This commit makes it so the position of the context menu doesn't require an additional render to apply so it fixes the flickering.

### DIFF
--- a/tools/devtools/src/devtools_app/store/store.js
+++ b/tools/devtools/src/devtools_app/store/store.js
@@ -14,8 +14,6 @@ export const store = reactive({
     componentsToggleBlacklist: new Set(),
   },
   contextMenu: {
-    top: 0,
-    left: 0,
     id: 0,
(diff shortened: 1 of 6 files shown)


tools/devtools/src/devtools_app/store/store.js
```diff
@@ -30,9 +28,9 @@ export const store = reactive({
       if (y + menuHeight > window.innerHeight) {
         y = window.innerHeight - menuHeight;
       }
-      this.left = x + "px";
-      // Need 25px offset because of the main navbar from the browser devtools
-      this.top = y - 25 + "px";
+      menu.style.left = x + "px";
+      // // Need 25px offset because of the main navbar from the browser devtools
```
ged-odoo, line 32: double comment... is it really necessary?

ged-odoo (approved):

PR #1442, opened 2023-05-24
[FIX] parser: make t-on stricter

### LINKED ISSUE #1441: `t-on` directive should be stricter
Currently the character after `t-on` which is generally a dash is ignored, this can result in weird bugs because `t-onclick` actually adds a listener for "lick" events.

[Playground](https://odoo.github.io/owl/playground/#eyJqcyI6ImltcG9ydCB7IENvbXBvbmVudCwgdXNlU3RhdGUsIG1vdW50LCB1c2VSZWYsIG9uTW91bnRlZCB9IGZyb20gXCJAb2Rvby9vd2xcIjtcbmNsYXNzIFJvb3QgZXh0ZW5kcyBDb21wb25lbnQge1xuICAgIHN0YXRpYyB0ZW1wbGF0ZSA9IFwiUm9vdFwiO1xuICAgIHNldHVwKCkge1xuICAgICAgICB0aGlzLnN0YXRlID0gdXNlU3RhdGUoe3ZhbDogMn0pO1xuICAgICAgICB0aGlzLnJvb3QgPSB1c2VSZWYoXCJyb290XCIpO1xuICAgICAgICBvbk1vdW50ZWQoKCkgPT4ge1xuICAgICAgICAgICAgdGhpcy5yb290LmVsLmRpc3BhdGNoRXZlbnQobmV3IEV2ZW50KFwibGlja1wiKSk7XG4gICAgICAgIH0pXG4gICAgfVxuICAgIHNldFZhbCgpIHtcbiAgICAgICAgdGhpcy5zdGF0ZS52YWwgPSAzO1xuICAgIH1cbn1cbm1vdW50KFJvb3QsIGRvY3VtZW50LmJvZHksIHsgdGVtcGxhdGVzOiBURU1QTEFURVMsIGRldjogdHJ1ZSB9KTtcbiIsImNzcyI6Ii5ncmVldGVyIHtcbiAgICBmb250LXNpemU6IDIwcHg7XG4gICAgd2lkdGg6IDMwMHB4O1xuICAgIGhlaWdodDogMTAwcHg7XG4gICAgbWFyZ2luOiA1cHg7XG4gICAgdGV4dC1hbGlnbjogY2VudGVyO1xuICAgIGxpbmUtaGVpZ2h0OiAxMDBweDtcbiAgICBiYWNrZ3JvdW5kLWNvbG9yOiAjZWVlZWVlO1xuICAgIHVzZXItc2VsZWN0OiBub25lO1xufVxuIiwieG1sIjoiPHRlbXBsYXRlcz5cbiAgPGRpdiB0LW5hbWU9XCJSb290XCIgdC1vbmNsaWNrPVwic2V0VmFsXCIgdC1lc2M9XCJzdGF0ZS52YWxcIiB0LXJlZj1cInJvb3RcIi8+XG48L3RlbXBsYXRlcz5cbiJ9)

ged-odoo:
Before this commit, the `t-on` directive assumed that the next character would be a -, and ignored it, so if someone would write `t-onclick` by mistake, Owl would then add an event handler on the `lick` event, instead of `click`.

With this commit, we improve the parser to make it stricter and fail if there is no dash.

closes #1441

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -336,10 +336,10 @@ function parseDOMNode(node: Element, ctx: ParsingContext): AST | null {
 
   for (let attr of nodeAttrsNames) {
     const value = node.getAttribute(attr)!;
-    if (attr.startsWith("t-on")) {
-      if (attr === "t-on") {
-        throw new OwlError("Missing event name with t-on directive");
-      }
+    if (attr === "t-on" || attr === "t-on-") {
+      throw new OwlError("Missing event name with t-on directive");
+    }
+    if (attr.startsWith("t-on-")) {
       on = on || {};
       on[attr.slice(5)] = value;
     } else if (attr.startsWith("t-model")) {
--- a/tests/compiler/parser.test.ts
+++ b/tests/compiler/parser.test.ts
@@ -1147,6 +1147,24 @@ describe("qweb parser", () => {
     });
   });
 
+  test("t-onclick without dash", async () => {
+    expect(() => parse(`<button t-onclick="add">Click</button>`)).toThrowError(
+      "Unknown QWeb directive: 't-onclick'"
+    );
+  });
+
+  test("t-on without event", async () => {
+    expect(() => parse(`<button t-on="add">Click</button>`)).toThrowError(
+      "Missing event name with t-on directive"
+    );
+  });
+
+  test("t-on- without event", async () => {
+    expect(() => parse(`<button t-on-="add">Click</button>`)).toThrowError(
+      "Missing event name with t-on directive"
+    );
+  });
+
   // ---------------------------------------------------------------------------
   // t-model
   // ---------------------------------------------------------------------------
@@ -1273,6 +1291,12 @@ describe("qweb parser", () => {
       on: { click: "someMethod" },
       slots: null,
     });
+  });
+
+  test("component with event handler", async () => {
+    expect(() => parse(`<MyComponent t-onclick="someMethod"/>`)).toThrowError(
+      "unsupported directive on Component: t-onclick"
+    );
   });
 
   test("component with t-ref", async () => {


ged-odoo:
@sdegueldre

src/compiler/parser.ts
```diff
@@ -336,10 +336,10 @@ function parseDOMNode(node: Element, ctx: ParsingContext): AST | null {
...
   for (let attr of nodeAttrsNames) {
     const value = node.getAttribute(attr)!;
-    if (attr.startsWith("t-on")) {
-      if (attr === "t-on") {
-        throw new OwlError("Missing event name with t-on directive");
-      }
+    if (attr === "t-on") {
+      throw new OwlError("Missing event name with t-on directive");
+    }
+    if (attr.startsWith("t-on-")) {
```
brboi, line 342: IMO:
- `t-on` should throw `"Unknown QWeb directive: 't-on'"`
- `t-on-` should throw `"Missing event name with t-on directive"`

```suggestion
    if (attr.startsWith("t-on-")) {
        const eventName = attr.slice(5);
        if(!eventName) {
            throw new OwlError("Missing event name with t-on directive");
        }
```

brboi: Maybe not, but with your version `t-on-` does not throw `"Missing event name with t-on directive"`

src/compiler/parser.ts
```diff
@@ -336,10 +336,10 @@ function parseDOMNode(node: Element, ctx: ParsingContext): AST | null {
 
   for (let attr of nodeAttrsNames) {
     const value = node.getAttribute(attr)!;
-    if (attr.startsWith("t-on")) {
-      if (attr === "t-on") {
-        throw new OwlError("Missing event name with t-on directive");
-      }
+    if (attr === "t-on" || attr === "t-on-") {
+      throw new OwlError("Missing event name with t-on directive");
+    }
```
brboi, line 341: What about these cases ? :-)
- `<button t-on-.prevent="someMethod"/>` 
- `<Component t-on-="someMethod"/>`
- `<Component t-on-.prevent="someMethod"/>` 
- `<t t-set-slot="someSlot" t-on-="someMethod" />`
- `<t t-set-slot="someSlot" t-on-.prevent="someMethod" />`
- `<t t-slot="someSlot" t-on-="someMethod" />`
- `<t t-slot="someSlot" t-on-.prevent="someMethod" />`

brboi (approved):

PR #1447, opened 2023-05-31
[FIX] hooks: useSubEnv and useChildSubEnv should apply on prototype chain

BastienFafchamps:
Before this commit:
When using useSubEnv or useChildSubEnv, only owned properties of the extensions are assigned to the new env.

After this commit:
Properties from the whole prototype chain of the extension are assigned to the env.

### DIFF
--- a/src/runtime/hooks.ts
+++ b/src/runtime/hooks.ts
@@ -36,8 +36,15 @@ export function useEnv<E extends Env>(): E {
 
 function extendEnv(currentEnv: Object, extension: Object): Object {
   const env = Object.create(currentEnv);
-  const descrs = Object.getOwnPropertyDescriptors(extension);
-  return Object.freeze(Object.defineProperties(env, descrs));
+  while (extension !== Object.prototype) {
+    const descrs = Object.getOwnPropertyDescriptors(extension);
+    Object.defineProperties(env, descrs);
+    for (const key in descrs) {
+      descrs[key].configurable = true;
+    }
+    extension = Object.getPrototypeOf(extension);
+  }
+  return Object.freeze(env);
 }
 
 /**
--- a/tests/components/__snapshots__/hooks.test.ts.snap
+++ b/tests/components/__snapshots__/hooks.test.ts.snap
@@ -361,6 +361,34 @@ exports[`hooks useRef hook: basic use 1`] = `
 }"
 `;
 
+exports[`hooks useSubEnv applies on prototype chain 1`] = `
+"function anonymous(app, bdom, helpers
+) {
+  let { text, createBlock, list, multi, html, toggler, comment } = bdom;
+  const comp1 = app.createComponent(\`Child\`, true, false, false, []);
+  
+  return function template(ctx, node, key = \\"\\") {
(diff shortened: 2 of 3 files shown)


src/runtime/hooks.ts
```diff
@@ -36,8 +36,15 @@ export function useEnv<E extends Env>(): E {
...
 function extendEnv(currentEnv: Object, extension: Object): Object {
   const env = Object.create(currentEnv);
-  const descrs = Object.getOwnPropertyDescriptors(extension);
-  return Object.freeze(Object.defineProperties(env, descrs));
+  while (extension !== Object.prototype) {
+    const descrs = Object.getOwnPropertyDescriptors(extension);
+    Object.defineProperties(env, descrs);
+    for (const key in descrs) {
+      descrs[key].configurable = true;
+    }
```
sdegueldre, line 44: This doesn't seem to do anything?

sdegueldre (commented):
Curious about the reason behind this change. It doesn't seem obvious to me that it should behave this way. The caller can flatten the object before calling use(Child)SubEnv if needed. This seems to encourage patterns that would lead to confusing behaviour, eg `useSubEnv(new SomeClass())` where the env will have all the properties and methods of SomeClass but will not be `instanceof SomeClass` and if the prototype of `SomeClass` is modified/patched the sub-env will not reflect the changes to the prototype.

src/runtime/hooks.ts
```diff
@@ -36,8 +36,15 @@ export function useEnv<E extends Env>(): E {
 
 function extendEnv(currentEnv: Object, extension: Object): Object {
   const env = Object.create(currentEnv);
-  const descrs = Object.getOwnPropertyDescriptors(extension);
-  return Object.freeze(Object.defineProperties(env, descrs));
+  while (extension !== Object.prototype) {
+    const descrs = Object.getOwnPropertyDescriptors(extension);
+    Object.defineProperties(env, descrs);
```
sdegueldre, line 41: Oh and because we're doing this while climbing the prototype chain, properties on the prototype will shadow properties on the object instead of the other way around which is undesirable in any case.

ged-odoo:
> Curious about the reason behind this change. It doesn't seem obvious to me that it should behave this way. The caller can flatten the object before calling use(Child)SubEnv if needed. 

Not easy to do in some cases. Here, we are in a situation where we want to do `useSubEnv(someOtherEnv)`, but if you flatten the other env, you evaluate immediately all the getters.

sdegueldre:
> if you flatten the other env, you evaluate immediately all the getters
Not if you flatten it in the same way as we do in this PR.

Doing `useSubEnv(otherEnv)` seems like a very weird use case and I'm curious how you even get in a situation where a component has a hold of 2 environments *and* wants an actual sub-env. It seems more likely you want to replace the env completely. I've had to do this in a few places in Odoo while migrating legacy code and did it by writing on owl internals `__owl__.env` and `__owl__.childEnv`. If this is a use case we want to support I'd rather be in favour of a hook that lets you replace the env wholesale instead.

Issue #1446, opened 2023-05-31
throw type error: setting getter-only property, in capture function


goyier:
i want to patch list renderer with property which only have getter,this error is occured.
![深度截图_选择区域_20230531145547](https://github.com/odoo/owl/assets/8132937/93cfbed5-cb67-4ee9-acc1-8a2b4577c407)
![深度截图_选择区域_20230531145601](https://github.com/odoo/owl/assets/8132937/a5965eb7-f73e-451a-b248-0334caddd489)
![深度截图_选择区域_20230531144549](https://github.com/odoo/owl/assets/8132937/156fc369-caa5-47d8-8379-73630ed8835b)

goyier:
![深度截图_选择区域_20230531145547](https://github.com/odoo/owl/assets/8132937/2f7f6ec2-2613-48f3-8c2d-f30ec9079409)
![深度截图_选择区域_20230531145601](https://github.com/odoo/owl/assets/8132937/72ceff59-129a-4561-9f45-a7294270c6cb)
![深度截图_选择区域_20230531144549](https://github.com/odoo/owl/assets/8132937/147d5c17-7e2b-45a9-bd40-710aae5814a5)

goyier:
the version of owl is 2.1.3

sdegueldre:
@goyier hello, while I have a fix pending for this problem, the fix in question may have pretty impactful performance consequences so we're still debating whether to merge it or not, so it might take some time.

In the meantime, I can suggest a workaround, which is to make the getter non-enumerable, while this ugly, it should fix your issue:

```js
const extraMethods = {
    // Non getter methods
}

Object.defineProperties(extraMethods, {
    showWSB: {
        get() {
            return this.wsbCode(); //...
        }
    },
    otherGetter: {
        get() {
            return this.otherValue;
        }
    }
});

patch(ListRenderer.prototype, 'patchName', extraMethods);
```

sdegueldre:
Because of the performance penalty introduced by the potential fix (https://github.com/odoo/owl/pull/1452), and the fact that it is in fact quite difficult to get an enumerable getter in a component, I'm closing this and suggest that anyone encountering this issue works around this by making their getters non-enumerable. This is the approach that we will take with the patch function in Odoo. Feel free to comment on this issue if you encounter another case with this problem and the suggested workaround does not work or is too annoying and we may consider reopening it.

PR #1452, opened 2023-06-08
[FIX] runtime: do not crash when capturing context with getter

sdegueldre:
When attempting to capture a rendering context that contains an enumerable getter, there is a crash because we attempt to write a value on the getter. This commit fixes that by manually climbing the prototype chain to copy the values instead, and ignoring getters.

closes https://github.com/odoo/owl/issues/1446


sdegueldre:
see https://github.com/odoo/owl/issues/1446#issuecomment-1643360546

PR #1463, opened 2023-06-23
[IMP] devtools: present observed keys in a more compact way

sdegueldre:
Previously, the keys were displayed separately from the target, on top of being displayed in bold inside the object when unfolded. This is redundant, and in most cases you have to unfold the object anyway because there are more keys observed than can be displayed.

The only "key" that cannot be displayed in bold inside the object is the "key changes" magic key. This commit replaces it by a small +/- badge on the right of the object's short content.

This commit also stops displaying observed object that are nested inside other observed objects at the top level, as it can be confusing, and it also heuristically gives a name to observed objects: if the observed object is a property of the component or one of its props it will be named accordingly, if not it will be named `[unknown]` (which may happen when reobserving reactive objects inside a service or inside the env, but is pretty rare in practice).

## Before:
image
Notice how nested keys were not highlighted even if observed:
image

## After:
image


tools/devtools/src/devtools_app/devtools_window/components_tab/details_window/subscriptions/subscriptions.xml
juliusc2066, line 1: You can probably delete the subscriptions component entirely at this point

Issue #1464, opened 2023-06-23
Devtools: handling of symbols in objects


sdegueldre:
Symbols don't show up in the short representation of objects, and objects are not expandable when they only contain symbol keys

PR #1468, opened 2023-06-29
[FIX] devtools: fix symbols handling and display

juliusc2066:
This commit fixes the three following issues:
- Symbols could never appear in shortened display of objects
- Objects which contained only symbols as keys would be considered to be empty and therefore not expandable
- Symbol value edition would create a new property on the object with the stringified symbol as key instead of updating its value 

closes https://github.com/odoo/owl/issues/1464

### DIFF
--- a/tools/devtools/src/page_scripts/owl_devtools_global_hook.js
+++ b/tools/devtools/src/page_scripts/owl_devtools_global_hook.js
@@ -117,6 +117,15 @@
             length += element.length;
             result.push(element);
           }
+          for (const key of Object.getOwnPropertySymbols(obj)) {
+            if (length > 25) {
+              result.push("...");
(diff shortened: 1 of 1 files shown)


tools/devtools/src/page_scripts/owl_devtools_global_hook.js
```diff
@@ -1433,8 +1443,13 @@
           return;
         }
       }
-      const key = path.pop().value;
+      let key = path.pop();
```
ged-odoo, line 1446: now, it's not really a key, is it? i propose `const elem = ...`
and then below:
```js
const key = elem.SymbolIndex ? ... : elem.value;
```

ged-odoo (approved):

Issue #1467, opened 2023-06-28
Devtools: handling of prototypes


juliusc2066:
The display of prototypes has issues when a non class object has a chain of prototypes. The devtools should display all the sub prototypes until the native JS one for generic Objects

PR #1471, opened 2023-07-03
[FIX] devtools: fix/imp env display

juliusc2066:
This commit first fixes how object prototype are detected so that it won't stop as soon as the constructor name of the object is "Object". This allows displaying every single prototype encountered and closes https://github.com/odoo/owl/issues/1467.

This commit also improves how the env of a component is displayed by expanding its chain of prototypes by default while keeping its keys lit as long as it is their first occurence in the chain.

### DIFF
--- a/tools/devtools/src/page_scripts/owl_devtools_global_hook.js
+++ b/tools/devtools/src/page_scripts/owl_devtools_global_hook.js
@@ -748,6 +748,9 @@
           child.contentType = "object";
           child.content = this.serializer.serializeItem(Object.getPrototypeOf(parentObj), true);
           child.hasChildren = true;
+          if (!oldTree && type === "env") {
+            child.toggled = true;
+          }
           break;
         case "set entries":
         case "map entries":
@@ -890,7 +893,7 @@
       const children = [];
       depth = depth + 1;
       let obj = this.getObjectProperty(path);
-      let oldBranch = this.getObjectInOldTree(oldTree, path, objType);
+      let oldBranch = oldTree && this.getObjectInOldTree(oldTree, path, objType);
       if (!obj) {
         return [];
       }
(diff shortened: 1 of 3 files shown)


tools/devtools/src/page_scripts/owl_devtools_global_hook.js
```diff
@@ -890,7 +893,10 @@
       const children = [];
       depth = depth + 1;
       let obj = this.getObjectProperty(path);
-      let oldBranch = this.getObjectInOldTree(oldTree, path, objType);
+      let oldBranch;
```
ged-odoo, line 896: usually, that kind of code is written like this:
```js
let oldBranch = oldTree && this.getObjectInOldTree(oldTree, path, objType);
```
or 
```js
let oldBranch = oldTree ? this.getObjectInOldTree(oldTree, path, objType) : null; // or false, or undefined, depending on your religious beliefs
```

tools/devtools/src/page_scripts/owl_devtools_global_hook.js
```diff
@@ -1061,14 +1067,14 @@
...
-      if (!(obj.constructor.name === "Object")) {
+      if (obj.__proto__) {
         prototype = this.serializeObjectChild(
           obj,
           { type: "prototype", childIndex: children.length },
           depth,
           objType,
           path,
-          oldBranch.children.at(-1),
+          oldBranch?.children.at(-1),
```
ged-odoo, line 1077: oh woah, i completely forgot that `.at` could take negative number

ged-odoo (commented):
looks ok to me. for the commit message, I think that you should just tag it as a `[FIX]`. One can argue that it is a functional fix :)

ged-odoo (approved):

Issue #1352, opened 2023-02-14
support t-foreach on an iterable object


mcm-odoo:
Can we support iterable objects in `t-foreach`?

```js
class MyIterable {
    *[Symbol.iterator]() {
        yield 1;
        yield 2;
        yield 3;
        // ...
    }
}
```

I also noticed that we don't support `t-foreach` on a `Set` or `Map`, they will also be supported as they are iterable.
`<t t-foreach="mySet" ...` where `mySet = new Set([1, 2, 3])` is valid but does not loop at the moment.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

PR #1476, opened 2023-07-12
[IMP] runtime: add support for Map and other iterables in t-foreach

sdegueldre:


src/runtime/template_helpers.ts
```diff
@@ -60,18 +60,24 @@ function withKey(elem: any, k: string) {
...
 
-function prepareList(collection: any): [any[], any[], number, any[]] {
-  let keys: any[];
-  let values: any[];
+function prepareList(collection: unknown): [unknown[], unknown[], number, undefined[]] {
+  let keys: unknown[];
+  let values: unknown[];
 
-  if (Array.isArray(collection)) {
+  if (!collection || typeof collection !== "object") {
```
ged-odoo, line 67: you can do this check in the else clause, so the usual case is (very very slightly) faster

tests/compiler/t_foreach.test.ts
```diff
@@ -105,6 +105,54 @@ describe("t-foreach", () => {
     expect(renderToString(template, context)).toBe(expected);
   });
 
+  test("iterate, Map param", () => {
+    const template = `
+        <div>
```
ged-odoo, line 110: not really important, but you can remove the `<div>` if you want. not really relevant to the test

tests/compiler/t_foreach.test.ts
```diff
@@ -105,6 +105,54 @@ describe("t-foreach", () => {
...
+        ["b", 2],
+        ["c", 3],
+      ]),
+    };
+    expect(renderToString(template, context)).toBe(expected);
+  });
+
+  test("iterate, Set param", () => {
+    const template = `
+        <div>
```
ged-odoo, line 128: same

tests/compiler/t_foreach.test.ts
```diff
@@ -193,7 +241,9 @@ describe("t-foreach", () => {
 
   test("throws error if invalid loop expression", () => {
     const test = `<div><t t-foreach="abc" t-as="item" t-key="item"><span t-key="item_index"/></t></div>`;
-    expect(() => renderToString(test)).toThrow("Invalid loop expression");
+    expect(() => renderToString(test)).toThrow(
+      'Invalid loop expression: "undefined" is not iterable'
```
ged-odoo, line 245: this is actually a pretty big improvement. nice

ged-odoo (commented):
don't forget to add `closes: #1352` in the commit message

ged-odoo (approved):

tests/compiler/t_foreach.test.ts
```diff
@@ -105,6 +105,64 @@ describe("t-foreach", () => {
...
+        [<t t-esc="item_index"/>: <t t-esc="item"/> <t t-esc="item_value"/>]
+      </t>`;
+    const expected = ` [0: 1 a]  [1: 2 b]  [2: 3 c] `;
+    const context = {
+      value: new Map([
+        ["a", 1],
+        ["b", 2],
+        ["c", 3],
+      ]),
+    };
```
seb-odoo, line 120: Hello,

I'm a bit confused with this test, I believe this should be "a 1", "b 2" and "c 3", as stated by the doc "for objects and maps, it provides the value (where `$as`  provides the key)"

It comes from
```ts
  } else if (collection instanceof Map) {
    keys = [...collection.keys()];
    values = [...collection.values()];
```

where keys and values are not "inverted", as opposed to what is done for the object case?

sdegueldre: Definitely a mistake, didn't realize the existing code for objects was inverted for some reason... I'll make a fix tomorrow.

sdegueldre: https://github.com/odoo/owl/pull/1502

PR #1484, opened 2023-07-19
[FIX] compiler: allow t-out on component tag

### LINKED ISSUE #1483: Unexpected different behaviour for t-out and t-esc when used on a component tag
Using a t-esc on a component tag does add the text to the component default slot, while using a t-out replaces the whole component tag by a text node. I think the behaviour should be the same for both.

```xml
<Comp t-esc="'foo'"/> => <div class="my-comp">foo</div>

<Comp t-out="'foo'"/> =>  "foo"
```

ged-odoo:
Before this commit, the template parser would allow using t-esc on a component tag (<MyComponent t-esc="expr"/>) but would incorrectly ignore the component when parsing a t-out: <MyComponent t-out="expr"/> would be parsed as <t t-out="expr"/>

This commit solves the issue, and also, moves the `t-out` parsing code next to `t-esc` so they have the same priority relatively to other directives.

closes #1483

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -235,10 +235,10 @@ function parseNode(node: Node, ctx: ParsingContext): AST | null {
     parseTCall(node, ctx) ||
     parseTCallBlock(node, ctx) ||
     parseTEscNode(node, ctx) ||
+    parseTOutNode(node, ctx) ||
     parseTKey(node, ctx) ||
     parseTTranslation(node, ctx) ||
     parseTSlot(node, ctx) ||
-    parseTOutNode(node, ctx) ||
     parseComponent(node, ctx) ||
     parseDOMNode(node, ctx) ||
     parseTSetNode(node, ctx) ||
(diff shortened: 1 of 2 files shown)


src/compiler/parser.ts
```diff
@@ -941,21 +938,23 @@ function normalizeTIf(el: Element) {
...
-    if (el.childNodes.length) {
-      throw new OwlError("Cannot have t-esc on a component that already has content");
-    }
-    const value = el.getAttribute("t-esc");
-    el.removeAttribute("t-esc");
-    const t = el.ownerDocument.createElement("t");
-    if (value != null) {
-      t.setAttribute("t-esc", value);
+function normalizeTEscTOut(el: Element) {
+  for (let tag of ["t-esc", "t-out"]) {
```
sdegueldre, line 942: It's a little weird to call an attribute "tag" (also you can use const because tag is never reassignedà

sdegueldre (approved):
Minor comment

PR #1491, opened 2023-07-20
[IMP] compiler: add support for t-for directive

sdegueldre:
This commit adds support for the `t-for` whose syntax and usage is similar to the syntax of the for..of loop in JS (and in fact it compiles to a for..of loop). This looping construct supports looping on arbitrary iterables and destructuring assignments, which would be difficult to support in a backward compatible manner on the existing t-foreach directive.


ged-odoo:
to think about for owl 3

sdegueldre:
Still on my wishlist but the code is incredibly out of date

PR #1525, opened 2023-09-29
[FIX] hooks: remove useEffect type circular dependency

LucasLefevre:
`useEffect` type is not compatible with later versions of Typescript.
(I tried with Typescript 5.2, I didn't check which specific
version had the breaking change)

This prevents other projects using owl (such as o-spreadsheet) to upgrade their own version of Typescript.

`<T extends [...T]>` raises the two following errors:

`Type parameter 'T' has a circular constraint.ts(2313)`
`A rest element type must be an array type.`

Also, JSDoc was not working as expected because the type `Effect` was never defined (with `@typedef`).
Hence `@param {Effect<T>} effect` didn't work.

### DIFF
--- a/tests/components/hooks.test.ts
+++ b/tests/components/hooks.test.ts
@@ -641,6 +641,56 @@ describe("hooks", () => {
       ]);
     });
 
+    test("effect types are inferred from dependencies", async () => {
+      // @ts-ignore (declared but never used)
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      class MyComponent extends Component {
+        static template = xml`<div/>`;
+
+        setup() {
+          useEffect(
+            (a, b) => {
+              expectType<number>(a);
+              expectType<string>(b);
+            },
+            () => [3, "hello"]
+          );
+        }
+      }
+    });
+
+    test("effect type allows an effect with partial dependencies parameters", async () => {
+      // @ts-ignore (declared but never used)
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      class MyComponent extends Component {
+        static template = xml`<div/>`;
+
+        setup() {
+          useEffect(
+            (a) => {
+              expectType<number>(a);
+            },
+            () => [3, "hello"]
+          );
+        }
+      }
+    });
+
+    test("effect type allows an effect with no dependency parameter", async () => {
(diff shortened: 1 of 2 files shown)


sdegueldre:
Did you check that type inference is preserved for the parameters?

Also you're breaking the ability to have an effect with dependencies whose effect executor does not take all dependencies as argument.

In general, prefer `unknown` over `any`. In this case it probably comes from the previous code but unless there's a good reason to use any, let's not.

LucasLefevre:
> Did you check that type inference is preserved for the parameters?

Yes I did check :)
I added a few tests (even though I'm not sure they are particularly useful, especially the last two. Let me know what you think)

> Also you're breaking the ability to have an effect with dependencies whose effect executor does not take all dependencies as argument.

Ah yes true. TS infers the generic type in the other way my brain was :sweat_smile: 
Funnily enough, my previous attempt did work with some parameters, but was broken with no parameters at all. I restored `EffectDeps`. Anyway, it's better to not depend on TS inference implementation since I don't think it's standardized (/documented).

```ts
// worked with previous version
useEffect(
  (a) => {},
  () => [3, "hello"]
);
// that was broken with my previous attempt
useEffect(
  () => {},
  () => [3, "hello"]
);
```

> In general, prefer unknown over any. In this case it probably comes from the previous code but unless there's a good reason to use any, let's not.

True

Other comments:
- I removed the attempt to make JSDoc work. The build process removes my `@typedef` comments, I don't have time to investigate right now.
- 3 (unrelated) tests are failing when I run the test suite locally, I don't know if it's known.

tests/components/hooks.test.ts
```diff
@@ -641,6 +641,56 @@ describe("hooks", () => {
       ]);
     });
 
+    test("effect types are inferred from dependencies", async () => {
+      // @ts-ignore (declared but never used)
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      class MyComponent extends Component {
```
LucasLefevre, line 647: `@ts-ignore` required for jest
`// eslint-disable-next-line` required for linting
Don't know if there is another way

sdegueldre:
Seems fine. Not sure how I feel about tests for types since these will break the test suite even if they're skipped but I guess we can deal with that if and when it happens, thanks for the changes!

PR #1527, opened 2023-09-29
[FIX] doc: onWillDestroy mistasken for onWillUnmount

cammarosano:
Considering the context, the writer probably meant "onWillDestroy" instead of "onWillUnmount" here.


sdegueldre (approved):
Good catch, thank you

PR #1528, opened 2023-09-29
[FIX] doc: error handling example

cammarosano:
A few corrections to make the example more realistic:
- The component has no `error` property, so the writer probably meant `state.error`.
- The "content" slot is not being set, and "default" does the job in this case.
- The closing tag was missing in the `ErrorBoundary`'s template.


sdegueldre (approved):

sdegueldre:
Thanks for the corrections!

PR #1545, opened 2023-10-19
[ADD] owl-vision: vscode extension initial commit

BastienFafchamps:
Owl Vision is a vscode extension that improves owl developpement by adding
syntax highlights in templates and commands to easly navigate between
components and templates.

It also adds a Component snippet.

The following commands have been added:

* `Owl Vision: Find Template`:
    - If the cursor is on a template name, finds the corresponding template.
    - If the cursor is on a component, finds the template of the selected component.
* `Owl Vision: Find Component`: Finds the selected component definition.
* `Owl Vision: Switch`: Finds the corresponding template or component
    depending on the current file.
* `Owl Vision: Switch Besides`: Finds the corresponding template or component
    depending on the current file and opens it besides.

The following settings have been added to make the extension more configurable:

* `owl-vision.include`: Glob filter for files to include while searching.
* `owl-vision.exclude`: Glob filter for files to exclude while searching.

The .gitignore files has also been updated to accommodate to vscode extension development.


tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+		return undefined;
+	}
+
+	public async findTemplate(templateName: string): Promise<SearchResult | undefined> {
+		const isComponentName = templateName.match(/^[A-Z][a-zA-Z0-9_]*$/);
+
+		if (isComponentName) {
+			const componentResult = await this.findComponent(templateName);
+			if (!componentResult) {
+				return undefined;
```
sdegueldre, line 129: nit: prefer a bare return over `return undefined`

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+		return undefined;
+	}
+
+	private getTemplateNameInJS(str: string): string | undefined {
+		const matches = new RegExp(`template\\s*=\\s*["']([a-zA-Z0-9_\\-\\.]+)["']`, 'g').exec(str);
+		if (matches) {
+			return matches[1];
+		} else {
+			return undefined;
+		}
```
sdegueldre, line 170: useless else branch

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	private async findComponentFromTemplateName(templateName: string): Promise<SearchResult | undefined> {
+		const result = await this.find(templateName, `template\\s*=\\s*["'](${templateName})["']`, this.componentCache, "js");
+		if (result) {
+			return result;
+		}
+
+		vscode.window.showWarningMessage(`Could not find a component for "${templateName}"`);
+		return undefined;
```
sdegueldre, line 161: useless return

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	private async getFiles(
+		searchQuery: string,
+		include: vscode.GlobPattern,
+		exclude: vscode.GlobPattern,
+	): Promise<Array<vscode.Uri>> {
+		const files = await vscode.workspace.findFiles(include, exclude);
+		const parts = searchQuery.split(".").flatMap(s => s.split(/\.?(?=[A-Z])/)).map(s => s.toLowerCase());
```
sdegueldre, line 212: the `\.?` in the second split is useless: you've already split on the dot so the parts will never contain more dots.

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+		if (cache.has(name)) {
+			return cache.get(name);
+		}
+
+		const include = `{${vscode.workspace.getConfiguration().get(`owl-vision.${fileType}.include`)}}`;
+		const exclude = `{${vscode.workspace.getConfiguration().get(`owl-vision.${fileType}.exclude`)}}`;
+		const files = await this.getFiles(name, include, exclude);
+
+		for (const file of files) {
+			let result = await this.findInFile(file, new RegExp(searchQuery, 'g'), 6);
```
sdegueldre, line 197: Could be `const`. Where does the magic 6 come from? Seems like it sort of works because `class ` and `t-name` are both 6 characters, but this causes the cursor to be outside the quotes for t-name and in the incorrect position if there is more than one space after the class declaration (`class     MyComponent`). The cursor is also in a very weird place when using switch/switchBesides. If you build your regexes to use lookbehind, you won't need an offset at all and it will probably work better in all cases

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+				if (filepath.includes(part)) {
+					score++;
+				}
+			}
+			if (filepath.includes("legacy")) {
+				score = score / 2.0;
+			}
+			return { score, file };
+		}).sort((a, b) => {
+			return a.score > b.score ? -1 : 1;
```
sdegueldre, line 231:
```suggestion
		}).sort((a, b) => b.score - a.score)
```

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+			let score = 0;
+			if (path.dirname(filepath) === currentDir) {
+				score += 99;
+			}
+			for (const part of parts) {
+				if (filepath.includes(part)) {
+					score++;
+				}
+			}
+			if (filepath.includes("legacy")) {
```
sdegueldre, line 226: Seems very "Odoo 16" specific

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+
+	private async findInFile(
+		file: vscode.Uri,
+		searchQuery: RegExp,
+		selectionStartOffset: number,
+	): Promise<SearchResult | undefined> {
+		const document = await vscode.workspace.openTextDocument(file);
+		const text = document.getText();
+		const regex = searchQuery;
+		const match = regex.exec(text);
```
sdegueldre, line 245: `RegExp.prototype.exec` is a footgun because it mutates the regexp object (which this function doesn't own), prefer `String.prototype.match(All)` (you'll need matchAll if your regexp are global and you want the indices of the match) or `String.prototype.search` instead

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+		if (match !== null) {
+			return new SearchResult(file, new vscode.Range(
+				document.positionAt(match.index + selectionStartOffset),
+				document.positionAt(match.index + selectionStartOffset)
+			));
+		}
+		return undefined;
```
sdegueldre, line 252: useless return

tools/owl-vision/src/utils.ts
```diff
@@ -0,0 +1,39 @@
+import * as vscode from 'vscode';
+
+let statusMessage: any = undefined;
```
sdegueldre, line 3: should probably not be any. Maybe `ReturnType<vscode.window["createStatusBarItem"]>`?

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+		if (result) {
+			return result;
+		}
+
+		vscode.window.showWarningMessage(`Could not find a component for "${templateName}"`);
+		return undefined;
+	}
+
+	private getTemplateNameInJS(str: string): string | undefined {
+		const matches = new RegExp(`template\\s*=\\s*["']([a-zA-Z0-9_\\-\\.]+)["']`, 'g').exec(str);
```
sdegueldre, line 165: Prefer regexp literal + match/search

sdegueldre (approved):
Most of these comments are things that can be adressed later, the one thing that I'd like to be fixed before merging is the indents being tabs as this would generate a lot of diff in a subsequent PR.

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+		if (matches) {
+			return matches[1];
+		} else {
+			return undefined;
+		}
+	}
+
+	private async find(
+		name: string,
+		searchQuery: string,
```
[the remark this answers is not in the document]
BastienFafchamps: I'm not sure that what we want, all the regex made when calling this function have variables that needs to be inserted in the regex and the only way I think we can do it is by building the regex using strings.
I still changed other string regex to RegExp where I could though.

tools/owl-vision/src/search.ts
```diff
@@ -0,0 +1,255 @@
...
+		cache: Map<string, SearchResult>,
+		fileType: "js" | "xml",
+	) {
+		if (cache.has(name)) {
+			return cache.get(name);
```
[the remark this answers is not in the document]
BastienFafchamps: This issue seems somewhat context to fix now, but I'm now only searching through a limited amount of files and it seems to be working consistently and fast enough. I'll remove the caching logic for now, we might add it back later if needed.

ged-odoo (approved):

PR #1560, opened 2023-11-22
[IMP] types: correctly support `Function` type for props-validation

### LINKED ISSUE #1448: BaseType in props validation is missing type `FunctionConstructor`

sdegueldre:
Previously, having a Function as a type in the static props description of a component would only work if the component was not the root component, as the static props description on Component was "any", whereas the static props description on ComponentConstructor was "Schema". This meant that static props description on non-root components was not type-checked, and on root components it was type-checked only on the mount call.

This commit makes it so that static type description is of type "Schema" on Component, now causing static props description to be type-checked, and adds `typeof Function` to the `BaseType` union, which allows declaring that a component expects a function as a prop.

Closes #1448

### DIFF
--- a/src/runtime/component.ts
+++ b/src/runtime/component.ts
@@ -23,7 +23,7 @@ export type ComponentConstructor<P extends Props = any, E = any> = (new (
 
 export class Component<Props = any, Env = any> {
   static template: string = "";
-  static props?: any;
+  static props?: Schema;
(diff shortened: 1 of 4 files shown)


sdegueldre:
cc @hokolomopo

ged-odoo (approved):

PR #1580, opened 2024-01-10
[IMP] template set config: getTemplate function

Polymorphe57:
A template set can now obtain a template by calling the function
getTemplate (if any) received in the initial config.
The returned value can be an Element instance, a template string, a
function (i.e. a compiled template), or null. In the last case, owl will
look into the template set rawTemplates object.

### DIFF
--- a/src/runtime/template_set.ts
+++ b/src/runtime/template_set.ts
@@ -13,6 +13,7 @@ export interface TemplateSetConfig {
   translatableAttributes?: string[];
   translateFn?: (s: string) => string;
   templates?: string | Document | Record<string, string>;
+  getTemplate?: (s: string) => Element | Function | string | void;
 }
 
 export class TemplateSet {
@@ -22,6 +23,7 @@ export class TemplateSet {
   dev: boolean;
   rawTemplates: typeof globalTemplates = Object.create(globalTemplates);
   templates: { [name: string]: Template } = {};
+  getRawTemplate?: (s: string) => Element | Function | string | void;
   translateFn?: (s: string) => string;
   translatableAttributes?: string[];
   Portal = Portal;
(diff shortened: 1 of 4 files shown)


src/runtime/template_set.ts
```diff
@@ -13,6 +13,7 @@ export interface TemplateSetConfig {
   translatableAttributes?: string[];
   translateFn?: (s: string) => string;
   templates?: string | Document | Record<string, string>;
+  getTemplate?: (s: string) => (Element|string|null);
```
Polymorphe57, line 16: Element or something else?

Polymorphe57:
Hello @sdegueldre, could you please have a look?

tests/compiler/template_set.test.ts
```diff
@@ -78,4 +78,64 @@ describe("loading templates", () => {
     context.addTemplates(xml);
     expect(Object.keys(context.rawTemplates)).toEqual([]);
   });
+
+  test("function getTemplate in config", () => {
```
sdegueldre, line 82: Some names for these tests would be welcome

src/runtime/template_set.ts
sdegueldre: Element is fine.

src/runtime/template_set.ts
```diff
@@ -22,6 +23,7 @@ export class TemplateSet {
   dev: boolean;
   rawTemplates: typeof globalTemplates = Object.create(globalTemplates);
   templates: { [name: string]: Template } = {};
+  getRawTemplate?: (s: string) => Element | Function | string | null;
```
sdegueldre, line 26: I think returning undefined instead of null would make more sense, it allows avoiding an explicit null return.

ged-odoo:
dont forget to update the doc page on App

sdegueldre (approved):

PR #1581, opened 2024-01-12
[IMP] reactivity: replace sets with small arrays for performance

sdegueldre:
While Sets have better lookup complexity than arrays, because of the large constant factors, small arrays can perform better than small sets when checking for inclusion.

In practice, replacing both of the raw types sets with arrays can improve performance of reactive-heavy workloads by as much as 30%.

Considering the reactivity code is very hot when rendering data-heavy components, and the low impact on readability of the fix, the cost-benefit analysis is clearly in favour of making the fix.

### DIFF
--- a/src/runtime/reactivity.ts
+++ b/src/runtime/reactivity.ts
@@ -20,8 +20,9 @@ type CollectionRawType = "Set" | "Map" | "WeakMap";
 const objectToString = Object.prototype.toString;
 const objectHasOwnProperty = Object.prototype.hasOwnProperty;
 
-const SUPPORTED_RAW_TYPES = new Set(["Object", "Array", "Set", "Map", "WeakMap"]);
-const COLLECTION_RAWTYPES = new Set(["Set", "Map", "WeakMap"]);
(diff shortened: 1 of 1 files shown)


aab-odoo (approved):

src/runtime/reactivity.ts
```diff
@@ -45,7 +46,7 @@ function canBeMadeReactive(value: any): boolean {
   if (typeof value !== "object") {
     return false;
   }
-  return SUPPORTED_RAW_TYPES.has(rawType(value));
+  return SUPPORTED_RAW_TYPES.includes(rawType(value));
```
ged-odoo, line 49: at this point, maybe it's even faster to simply check for equality:
```js
const type = rawType(value);
return type === "Object" || type === "Array" || type === "Set" || type === "Map" || type === "WeakMap";
```
Same for collections. what do you think? @sdegueldre

sdegueldre: I tried with an inline switch previously when experimenting but perf was about the same, I'm guessing this is already very easy to JIT efficiently but may require more exploration.

PR #1594, opened 2024-03-11
[FIX] docs: code in example cannot run

haumenphai:


sdegueldre:
Good catch, thanks for the contribution!

PR #1599, opened 2024-03-26
[IMP] types: correctly support Set type for props-validation

pro-odoo:
Previously, having a Set as a type in the static props description of a component would would lead to a type error.

This commit adds Set to the list of supported types in the Schema


sdegueldre:
Superseded by https://github.com/odoo/owl/pull/1600

PR #1615, opened 2024-05-22
[IMP] index: export batched utility function

LucasLefevre:
'batched' will be used in odoo/o-spreadsheet.
There's also a copied version (slightly modified) in odoo/odoo.

It shows it could be useful outside owl.


ged-odoo (commented):
don't forget to update the doc (the function should now be explained and there is a page that explains all exported symbols)

ged-odoo (approved):

PR #1639, opened 2024-09-19
[IMP] make set of timeout-able hooks (and their timeouts)

xmo-odoo:
Also unnest the handling of `result` via guard clauses, and generate messages as close as possible to use site, keeping the error construction itself where it currently is as the goal is specifically to point back to the *definition* site for the hook function.


sdegueldre (approved):
Definitely a good bit cleaner than it was. Is the purpose to make these timeouts configurable at some point?

sdegueldre:
@ged-odoo I can't merge PRs on this repo anymore so this is probably for you

ged-odoo (approved):

ged-odoo:
yeah, @sdegueldre your r+ privilege is suspended until you gave us some interesting stories...

PR #1641, opened 2024-09-26
[IMP] owl: add basic support for sub roots

### LINKED ISSUE #1640: app should have a way to have multiple roots
Knowledge has to create a new app for each embedded components. This is slightly annoying because there is some bookkeeping to do to copy the app configuration. Also, the sub app cannot have the same template cache, since compiled templates keep a reference to the app and doing that may cause weird crashes.

Another usecase could be website, which may have 1 single owl app with multiple roots all over the page.

I don't think that it would be very difficult to modify the App class to support multiple root nodes:
```js
app.mount(Component, root)
```

ged-odoo:
In this commit, we extend the owl App class to support multiple sub roots. This is useful for situations where we want to mount sub components in non-managed DOM. This is exactly what the Knowledge app is doing, with mounting views in an html editor.

Currently, this requires some difficult and fragile hacks, and still, the result is that it is very easy to mix components from the main App and a SubApp.  But Knowledge does not actually care about creating a sub app. It only needs the possibility to mount sub components in dynamic places.

closes #1640

### DIFF
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -16,10 +16,13 @@ export interface Env {
   [key: string]: any;
 }
 
-export interface AppConfig<P, E> extends TemplateSetConfig {
-  name?: string;
+export interface RootConfig<P, E> {
   props?: P;
   env?: E;
+}
+
+export interface AppConfig<P, E> extends TemplateSetConfig, RootConfig<P, E> {
+  name?: string;
   test?: boolean;
   warnIfNoStaticProps?: boolean;
 }
@@ -49,6 +52,12 @@ declare global {
   }
 }
 
+interface Root<P, E> {
+  node: ComponentNode<P, E>;
+  mount(target: HTMLElement | ShadowRoot, options?: MountOptions): Promise<Component<P, E>>;
+  destroy(): void;
+}
+
 window.__OWL_DEVTOOLS__ ||= { apps, Fiber, RootFiber, toRaw, reactive };
 
 export class App<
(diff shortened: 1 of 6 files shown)


src/runtime/app.ts
```diff
@@ -101,6 +108,30 @@ export class App<
...
+    this.subRoots.add(node);
+    return {
+      root: node,
+      mount: (target: HTMLElement | ShadowRoot, options?: MountOptions) => {
+        App.validateTarget(target);
+        if (this.dev) {
+          validateProps(SubRoot, subProps, { __owl__: { app: this } });
+        }
+        const prom = this.mountNode(node, target, options);
+        return prom;
```
sdegueldre, line 125: Shouldn't mount call createRoot?

src/runtime/app.ts
```diff
@@ -134,6 +165,9 @@ export class App<
 
   destroy() {
     if (this.root) {
```
sdegueldre, line 167: Will knowledge have a main root? If not then destroy will not work in this current state

aab-odoo (approved):

src/runtime/app.ts
ged-odoo: yes, they will have to destroy the root manually using the return object from createRoot. The code here is just for correctness.

src/runtime/app.ts
ged-odoo: done (but not as elegant as one might expect, i had to remove something from the subroot sets. this could probably be improved in the future in owl 3. i think the App class is not designed properly, it should basically not have a mount method, only a createRoot.

sdegueldre (approved):

Issue #1650, opened 2024-11-06
add "user-space" directives


ged-odoo:
@jpp-odoo 
here is a problem: to detect the middle click operation on a htmlelement, we need 2 event handlers and a non trivial amount of code to make it work across browsers and os. This can look like this:
```xml
  <div t-on-click="handleClick" t-on-auxclick="handleAuxClick">...</div>
```
and implementing some code in js.

Now, an alternative could be to do it with refs: a `t-ref` in the template, and then a hook that would add the two event listeners and the js logic.  However, it does not work so nicely when we are working in the context of a list of elements (with `t-foreach`). We need dynamic refs, it's clunky and not really documented/exposed.

Another alternative would be to allow "user space" directives: an owl app could be configured with a list of additional "directives", which would actually be a kind of macro that hook into the xml before the compilation.

```js
  new App(..., {
    userDirectives: {
     middleclick: function (xmlnode, value) {}
   }
  });
```
this fonction could transform the xml element and add `t-on-click` and `t-on-auxclick` like this:
```xml
<div t-user-middleclick="onMiddleClick">...</div>
```
into 
```xml
<div t-on-click="processClick(onMiddleClick)" t-on-auxclick="processAuxClick(onMiddleClick)">...</div>
```

PR #1651, opened 2024-11-07
[IMP] parser: add support for user-space directives

jpp-odoo:
This commit adds the support for user-space directives. To use the user-space directive, an Object of functions needs to be configured on the owl APP:
```js
 new App(..., {
    userDirectives: {
     my-user-directive: function (el, value) {
            el.setAttribute("t-on-click", value);
            return el;
       }
   }
  });
```
The functions will be called when a user directive with the name of the function is found. The original element will be replaced with the one returned by the function.
This :
```xml
<div t-user-my-user-directive="click" />
```
will be replace by :
```xml
<div t-on-click="value"/>
```

issue : https://github.com/odoo/owl/issues/1650

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -1,4 +1,5 @@
 import { OwlError } from "../common/owl_error";
+import type { customDirectives } from "../common/types";
 import { parseXML } from "../common/utils";
 
 // -----------------------------------------------------------------------------
@@ -198,23 +199,26 @@ export type AST =
 // -----------------------------------------------------------------------------
 const cache: WeakMap<Element, AST> = new WeakMap();
 
-export function parse(xml: string | Element): AST {
+export function parse(xml: string | Element, customDir?: customDirectives): AST {
+  const ctx = {
+    inPreTag: false,
+    customDirectives: customDir,
+  };
   if (typeof xml === "string") {
     const elem = parseXML(`<t>${xml}</t>`).firstChild as Element;
-    return _parse(elem);
+    return _parse(elem, ctx);
   }
   let ast = cache.get(xml);
   if (!ast) {
     // we clone here the xml to prevent modifying it in place
-    ast = _parse(xml.cloneNode(true) as Element);
+    ast = _parse(xml.cloneNode(true) as Element, ctx);
     cache.set(xml, ast);
   }
   return ast;
 }
 
-function _parse(xml: Element): AST {
+function _parse(xml: Element, ctx: ParsingContext): AST {
   normalizeXML(xml);
-  const ctx = { inPreTag: false };
   return parseNode(xml, ctx) || { type: ASTType.Text, value: "" };
 }
 
@@ -222,6 +226,7 @@ interface ParsingContext {
   tModelInfo?: TModelInfo | null;
   nameSpace?: string;
   inPreTag: boolean;
+  customDirectives?: customDirectives;
 }
 
 function parseNode(node: Node, ctx: ParsingContext): AST | null {
@@ -229,6 +234,7 @@ function parseNode(node: Node, ctx: ParsingContext): AST | null {
     return parseTextCommentNode(node, ctx);
   }
   return (
+    parseTCustom(node, ctx) ||
     parseTDebugLog(node, ctx) ||
     parseTForEach(node, ctx) ||
     parseTIf(node, ctx) ||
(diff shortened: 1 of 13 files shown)


tests/app/app.test.ts
```diff
@@ -201,4 +201,22 @@ describe("app", () => {
...
   });
+
+  test("can add functions to the bdom", async () => {
+    const steps: string[] = [];
+    class SomeComponent extends Component {
+      static template = xml`<div t-on-click="() => __globals__.plop('click')" class="my-div"/>`;
+    }
+    const app = new App(SomeComponent, {
+      globalValues: {
+        plop: (string : any) => {
```
kebeclibre, line 212: not a fan of the API: calling `__globals__.plop()` is not natural.
1. Find a way at compile time to destructure that object on demand, a bit like helpers and bdom are.
or
2. rename every member of the globalValue to `custom[Value]`

Find a way to protect the helpers and __globals__ from user alteration: if user's code accesses via `__globals__.myfunc()` and myfunc is not an arrow function,  then myfunc could access, modify, add to the __globals__ object (via its `this`)
Also, it is obviously a side effect of that feature, but it is another door open to memory leaks and errors due to scoping things into arbitrary, unkillable closures

kebeclibre: What happens when:
- A "main" app (let's the one spawning the webclient) defines such globals
- a secondary app (lets say the ones spawned by the htmleditor) doesn't define such globals.
In the secondary app, it is legitimate to want to spawn the same components that the main app would.
Does the globals object gets shared, or do we expect a crash ?

ged-odoo: there is no secondary app, that's why we have owl roots. they all share the same config/templates/translation function/settings and now global and user directives. should be fine

sdegueldre: > it is another door open to memory leaks and errors due to scoping things into arbitrary, unkillable closures

I for one am just not seeing it, mind expanding on that?

src/compiler/parser.ts
```diff
@@ -197,8 +198,12 @@ export type AST =
 // Parser
 // -----------------------------------------------------------------------------
 const cache: WeakMap<Element, AST> = new WeakMap();
+let customDirectives: customDirectives = {};
 
-export function parse(xml: string | Element): AST {
+export function parse(xml: string | Element, customDir?: customDirectives): AST {
+  if (customDir) {
```
ged-odoo, line 204: this is really ugly. is there a reason not to use the parsing context? you can add information to it

src/compiler/parser.ts
```diff
@@ -277,6 +283,42 @@ function parseTextCommentNode(node: Node, ctx: ParsingContext): AST | null {
...
+      const modifier = attr.split(".").length > 1 ? attr.split(".")[1] : undefined;
+      node.removeAttribute(attr);
+      try {
+        node = customDirective(node, value, modifier);
+      } catch (error) {
+        throw new OwlError(
+          `Custom directive "${directiveName}" throw the following error: ${error}`
+        );
+      }
+      if (!(node instanceof Element)) {
```
ged-odoo, line 311: i am thinking that maybe, we want the allow a directive not to return anything (so it modifies in place the element)

src/runtime/template_set.ts
```diff
@@ -97,7 +108,8 @@ export class TemplateSet {
       this.templates[name] = function (context, parent) {
         return templates[name].call(this, context, parent);
       };
-      const template = templateFn(this, bdom, helpers);
+      const runtimeUtils = { ...helpers, __globals__: { ...this.globalValues } };
```
ged-odoo, line 111: not sure why you spread this.globalValues? is there a need to specifically do that to avoid modifying it? 
```js
      const runtimeUtils = { ...helpers, __globals__: this.globalValues };
```

tests/app/app.test.ts
kebeclibre: It was just a general observation. In the history of what we did in odoo, we realized progressively that when passing callbacks everywhere we take the risk to have some objects in a closure that can hold the whole OWL Tree. In this particular case, those closures would be, indeed unkillable. But I don't think they will hold much information......

src/runtime/template_set.ts
```diff
@@ -97,7 +104,8 @@ export class TemplateSet {
       this.templates[name] = function (context, parent) {
         return templates[name].call(this, context, parent);
       };
-      const template = templateFn(this, bdom, helpers);
+      const runtimeUtils = { ...helpers, __globals__: this.globalValues };
```
ged-odoo, line 107: this is done everytime a component is instantiated. it would be cheaper to do it just once in the setup. so, remove this line

src/runtime/template_set.ts
```diff
@@ -97,7 +104,8 @@ export class TemplateSet {
...
+      const template = templateFn(this, bdom, runtimeUtils);
```
ged-odoo, line 108:
```js
      const template = templateFn(this, bdom, this.runtimeUtils);
```

ged-odoo (approved):

PR #1654, opened 2024-11-25
[FIX] compiler: multiple modifiers on t-custom

jpp-odoo:
Since [1] it's possible to have a custom directive, this directive only supported one modifier. This commit, add the support to have multiple modifiers on a t-custom directive.

[1] : https://github.com/odoo/owl/commit/7e687234bf40bdcf5e6598401d8604d66b14ad32

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -299,10 +299,10 @@ function parseTCustom(node: Element, ctx: ParsingContext): AST | null {
         throw new OwlError(`Custom directive "${directiveName}" is not defined`);
       }
       const value = node.getAttribute(attr)!;
-      const modifier = attr.split(".").length > 1 ? attr.split(".")[1] : undefined;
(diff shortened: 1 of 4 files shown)


src/compiler/parser.ts
```diff
@@ -299,10 +299,10 @@ function parseTCustom(node: Element, ctx: ParsingContext): AST | null {
         throw new OwlError(`Custom directive "${directiveName}" is not defined`);
       }
       const value = node.getAttribute(attr)!;
-      const modifier = attr.split(".").length > 1 ? attr.split(".")[1] : undefined;
+      const modifiers = attr.split(".").length > 1 ? attr.split(".").slice(1) : undefined;
```
ged-odoo, line 302: i think that if there are no modifiers, we should receive an empty list

ged-odoo (approved):

PR #1662, opened 2025-01-07
[IMP] compiler: add prop suffix name in the error message

rdeodoo:
Without this, you only get the following error:
```
UncaughtPromiseError > OwlError
Promesse non interceptée > Invalid prop suffix
OwlError: Invalid prop suffix
```

Which is not helping much to find the problematic code.


rdeodoo:
@ged-odoo This is a tiny step but I thought it might be helpful, that would have helped me at least. I can't add a reviewer myself on the PR.

sdegueldre (approved):

PR #1668, opened 2025-02-21
[FIX] code generator: prevent AST change

mcm-odoo:
This commit removes an AST change during the code generation of slots. Before, the `compileTSlot` function deleted the `t-props` attribute directly on `ast.attrs`. This creates wrong code when compiling a second time as the `t-props` attribute does not exist anymore.

### DIFF
--- a/src/compiler/code_generator.ts
+++ b/src/compiler/code_generator.ts
@@ -1350,17 +1350,16 @@ export class CodeGenerator {
       isMultiple = isMultiple || this.slotNames.has(ast.name);
       this.slotNames.add(ast.name);
     }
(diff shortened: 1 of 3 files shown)


ged-odoo:
not sure, but would it be possible to make a test? something like compiling twice a template?

sdegueldre:
Curious about one thing: when would we compile a parsed template more than once?

ged-odoo:
@sdegueldre it comes from a cache in owl to keep ast for the parser. it was added for spreadsheet, because it helps them a lot to speed their test suite. we usually don't mutate the ast in place, so i think it's not a common issue.

sdegueldre (approved):

ged-odoo:
@aab-odoo done. sorry

PR #1670, opened 2025-02-28
[IMP] runtime: add markup tag function

oomsveta:
Allows markup to be called as a tag function. The interpolated strings are then safely escaped for injection in HTML code.

Example usage:
```js
const maliciousInput = "<script>alert('💥💥')</script>";
const value = markup`<b>${maliciousInput}</b>`;
// no problem, maliciousInput is properly escaped
```

### DIFF
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -81,10 +81,50 @@ export async function loadFile(url: string): Promise<string> {
  */
 export class Markup extends String {}
 
+function _escapeHtml(str: any): string | Markup {
+  if (str instanceof Markup) {
+    return str;
+  }
+  if (str === undefined) {
+    return "";
+  }
+  if (typeof str === "number") {
+    return String(str);
+  }
+  [
+    ["&", "&amp;"],
+    ["<", "&lt;"],
+    [">", "&gt;"],
+    ["'", "&#x27;"],
+    ['"', "&quot;"],
+    ["`", "&#x60;"],
+  ].forEach((pairs) => {
+    str = String(str).replace(new RegExp(pairs[0], "g"), pairs[1]);
+  });
+  return str;
+}
+
 /*
  * Marks a value as safe, that is, a value that can be injected as HTML directly.
  * It should be used to wrap the value passed to a t-out directive to allow a raw rendering.
+ *
+ * If called as a tag function, the interpolated strings are escaped.
  */
-export function markup(value: any) {
-  return new Markup(value);
+export function markup(strings: TemplateStringsArray, ...placeholders: unknown[]): Markup;
+export function markup(value: string): Markup;
+export function markup(
+  valueOrStrings: string | TemplateStringsArray,
+  ...placeholders: unknown[]
+): Markup {
+  if (!Array.isArray(valueOrStrings)) {
+    return new Markup(valueOrStrings);
+  }
+  const strings = valueOrStrings;
+  let acc = "";
+  let i = 0;
+  for (; i < placeholders.length; ++i) {
+    acc += strings[i] + _escapeHtml(placeholders[i]);
+  }
+  acc += strings[i];
+  return new Markup(acc);
 }
--- a/tests/utils.test.ts
+++ b/tests/utils.test.ts
@@ -1,4 +1,4 @@
-import { batched, EventBus } from "../src/runtime/utils";
+import { batched, EventBus, markup } from "../src/runtime/utils";
 import { nextMicroTick } from "./helpers";
 
 describe("event bus behaviour", () => {
@@ -69,5 +69,35 @@ describe("batched", () => {
     expect(n).toBe(2);
     await nextMicroTick();
     expect(n).toBe(2);
+  });
+});
+
+const Markup = markup("").constructor;
+describe("markup", () => {
+  test("string is flagged as safe", () => {
+    const html = markup("<blink>Hello</blink>");
+    expect(html).toBeInstanceOf(Markup);
+  });
+  describe("tag function", () => {
+    test("interpolated values are escaped", () => {
+      const maliciousInput = "<script>alert('💥💥')</script>";
+      const html = markup`<b>${maliciousInput}</b>`;
+      expect(html.toString()).toBe("<b>&lt;script&gt;alert(&#x27;💥💥&#x27;)&lt;/script&gt;</b>");
+      expect(html).toBeInstanceOf(Markup);
+    });
+    test("interpolated markups aren't escaped", () => {
+      const shouldBeEscaped = "<script>alert('should be escaped')</script>";
+      const shouldnt = markup("<b>this is safe</b>");
+      const html = markup`<div>${shouldBeEscaped} ${shouldnt}</div>`;
+      expect(html.toString()).toBe(
+        "<div>&lt;script&gt;alert(&#x27;should be escaped&#x27;)&lt;/script&gt; <b>this is safe</b></div>"
+      );
+      expect(html).toBeInstanceOf(Markup);
+    });
+    test("quotes in interpolated values are escaped", () => {
+      const imgUrl = `lol" onerror="alert('xss')`;
+      const html = markup`<img src="${imgUrl}">`;
+      expect(html.toString()).toBe(`<img src="lol&quot; onerror=&quot;alert(&#x27;xss&#x27;)">`);
+    });
   });
 });
--- a/doc/reference/templates.md
+++ b/doc/reference/templates.md
@@ -193,6 +193,15 @@ The first `t-out` will act as a `t-esc` directive, which means that the content
 of `value1` will be escaped. However, since `value2` has been tagged as a markup,
 this will be injected as html.
 
+`markup` can also be used as a tag function, allowing the interpolated values to
+be safely escaped:
+
+```js
+const maliciousInput = "<script>alert('💥💥')</script>";
+// <b>&lt;script&gt;alert(&#x27;💥💥&#x27;)&lt;/script&gt;</b>
+const value = markup`<b>${maliciousInput}</b>`;
+```
+
 ### Setting Variables
 
 QWeb allows creating variables from within the template, to memoize a computation (to use it multiple times), give a piece of data a clearer name, ...


src/runtime/utils.ts
```diff
@@ -81,10 +81,44 @@ export async function loadFile(url: string): Promise<string> {
  */
 export class Markup extends String {}
 
+function _escapeHtml(str: any): string {
```
kebeclibre, line 84: I suppose this is done elsewhere in the code (see the code that compiles the t-set instruction).
Node.textContent is safe against XSS (https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) and is basically used in owl natively.

src/runtime/utils.ts
```diff
@@ -81,10 +81,44 @@ export async function loadFile(url: string): Promise<string> {
...
+export function markup(value: string): Markup;
+export function markup(valueOrStrings: string | TemplateStringsArray, ...placeholders: unknown[]): Markup {
+  if (!Array.isArray(valueOrStrings)) {
+    return new Markup(valueOrStrings);
+  }
+  const strings = valueOrStrings;
+  let acc = "";
+  let i = 0;
+  for (; i < placeholders.length; ++i) {
+    acc += strings[i] + _escapeHtml(placeholders[i]);
```
kebeclibre, line 120: should support placeholder instance if Markup

src/runtime/utils.ts
```diff
@@ -81,10 +81,44 @@ export async function loadFile(url: string): Promise<string> {
...
  * It should be used to wrap the value passed to a t-out directive to allow a raw rendering.
+ *
+ * If called as a tag function, the interpolated strings are escaped.
  */
-export function markup(value: any) {
-  return new Markup(value);
+export function markup(strings: TemplateStringsArray, ...placeholders: unknown[]): Markup;
+export function markup(value: string): Markup;
+export function markup(valueOrStrings: string | TemplateStringsArray, ...placeholders: unknown[]): Markup {
+  if (!Array.isArray(valueOrStrings)) {
```
kebeclibre, line 113: I think the way to go is to use the string.raw function https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw

I'm not really familiar with those apis.

kebeclibre (commented):
let's see what @jum-odoo thinks.

src/runtime/utils.ts
kebeclibre:
```js
el = document.createElement("span")
// <span>

el.textContent = `<img src="lol" onerror="alert('xss')">`
// `<img src="lol" onerror="alert('xss')">`

el.innerHTML
// `&lt;img src="lol" onerror="alert('xss')"&gt;` 
```

tests/utils.test.ts
```diff
@@ -71,3 +71,17 @@ describe("batched", () => {
...
+  test("string is flagged as safe", () => {
+    const html = markup("<blink>Hello</blink>");
+    expect(html).toBeInstanceOf(Markup);
+  });
+  test("when used as a tag function, interpolated values are escaped", () => {
+    const maliciousInput = "<script>alert('💥💥')</script>";
+    const html = markup`<b>${maliciousInput}</b>`;
+    expect(html.toString()).toBe("<b>&lt;script&gt;alert(&#x27;💥💥&#x27;)&lt;/script&gt;</b>");
+    expect(html).toBeInstanceOf(Markup);
+  });
```
sdegueldre, line 86: Would be good to add a test to check that interpolated markup is not escaped

seb-odoo:
markup-aware escape should be exported so we can use it in Odoo in place of `htmlEscape` that serves the exact same purpose.

See https://github.com/odoo/odoo/pull/199300 for a proof of concept in Odoo of this PR.

src/runtime/utils.ts
```diff
@@ -81,10 +81,50 @@ export async function loadFile(url: string): Promise<string> {
...
+    return new Markup(valueOrStrings);
+  }
+  const strings = valueOrStrings;
+  let acc = "";
+  let i = 0;
+  for (; i < placeholders.length; ++i) {
+    acc += strings[i] + _escapeHtml(placeholders[i]);
+  }
+  acc += strings[i];
```
seb-odoo, line 128: `return strings.reduce((res, str, i) => htmlJoin([res, markup(str), values[i]]), "");`

ged-odoo: random review: i prefer the for loop. this is more efficient than the proposed code (which adds a bunch of funciton calls/iterating over lists/allocating/...  In low level code like this, performance is more important than style

seb-odoo: Fair point, I didn't benchmark either but whatever is the fastest is fine!

seb-odoo: I removed `htmlJoin` which was overhead, but using `reduce` appears much faster than for loop (2x faster on firefox, 6x faster on chrome).

`return strings.reduce((res, str, i) => res + str + htmlEscape(values[i]))`

image

[benchmark](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJPQWpPWmlNRTJjWjUwazI5cmdyRHkiLCJjb2RlIjoiY29uc3QgdGV4dCA9IFwiYSB0ZXh0XCI7XG5jb25zdCBodG1sID0gREFUQS5tYXJrdXAoXCI8c3Bhbj50ZXN0PC9zcGFuPlwiKTtcbmZvciAobGV0IGkgPSAwOyBpIDwgMTAwMDAwOyBpKyspIHtcbiAgICBEQVRBLm1hcmt1cGA8ZGl2PiR7dGV4dH0gYW5kIGFsc28gJHtodG1sfTwvZGl2PmBcbn0iLCJuYW1lIjoiZm9yIGxvb3AiLCJkZXBlbmRlbmNpZXMiOltdfSx7ImlkIjoiOHNYOHg5MTZmblVQX0F1MnQxdThLIiwiY29kZSI6ImNvbnN0IHRleHQgPSBcImEgdGV4dFwiO1xuY29uc3QgaHRtbCA9IERBVEEubWFya3VwKFwiPHNwYW4-dGVzdDwvc3Bhbj5cIik7XG5mb3IgKGxldCBpID0gMDsgaSA8IDEwMDAwMDsgaSsrKSB7XG4gICAgREFUQS5tYXJrdXAyYDxkaXY-JHt0ZXh0fSBhbmQgYWxzbyAke2h0bWx9PC9kaXY-YFxufSIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJyZWR1Y2UifV0sImNvbmZpZyI6eyJuYW1lIjoiQmFzaWMgZXhhbXBsZSIsInBhcmFsbGVsIjp0cnVlLCJnbG9iYWxUZXN0Q29uZmlnIjp7ImRlcGVuZGVuY2llcyI6W119LCJkYXRhQ29kZSI6ImNsYXNzIE1hcmt1cCBleHRlbmRzIFN0cmluZyB7fVxuLyoqXG4gKiBFc2NhcGVzIGEgc3RyaW5nIGZvciBIVE1MLlxuICpcbiAqIEBwYXJhbSB7c3RyaW5nIHwgbnVtYmVyfSBbc3RyXSB0aGUgc3RyaW5nIHRvIGVzY2FwZVxuICogQHJldHVybnMge3N0cmluZ30gYW4gZXNjYXBlZCBzdHJpbmdcbiAqL1xuZnVuY3Rpb24gZXNjYXBlKHN0cikge1xuICAgIGlmIChzdHIgPT09IHVuZGVmaW5lZCkge1xuICAgICAgICByZXR1cm4gXCJcIjtcbiAgICB9XG4gICAgaWYgKHR5cGVvZiBzdHIgPT09IFwibnVtYmVyXCIpIHtcbiAgICAgICAgcmV0dXJuIFN0cmluZyhzdHIpO1xuICAgIH1cbiAgICBbXG4gICAgICAgIFtcIiZcIiwgXCImYW1wO1wiXSxcbiAgICAgICAgW1wiPFwiLCBcIiZsdDtcIl0sXG4gICAgICAgIFtcIj5cIiwgXCImZ3Q7XCJdLFxuICAgICAgICBbXCInXCIsIFwiJiN4Mjc7XCJdLFxuICAgICAgICBbJ1wiJywgXCImcXVvdDtcIl0sXG4gICAgICAgIFtcImBcIiwgXCImI3g2MDtcIl0sXG4gICAgXS5mb3JFYWNoKChwYWlycykgPT4ge1xuICAgICAgICBzdHIgPSBTdHJpbmcoc3RyKS5yZXBsYWNlQWxsKHBhaXJzWzBdLCBwYWlyc1sxXSk7XG4gICAgfSk7XG4gICAgcmV0dXJuIHN0cjtcbn1cbmZ1bmN0aW9uIGh0bWxFc2NhcGUoY29udGVudCkge1xuICAgIHJldHVybiBjb250ZW50IGluc3RhbmNlb2YgTWFya3VwID8gY29udGVudCA6IG1hcmt1cChlc2NhcGUoY29udGVudCkpO1xufVxuZnVuY3Rpb24gbWFya3VwKHZhbHVlT3JTdHJpbmdzLCAuLi5wbGFjZWhvbGRlcnMpIHtcbiAgaWYgKCFBcnJheS5pc0FycmF5KHZhbHVlT3JTdHJpbmdzKSkge1xuICAgIHJldHVybiBuZXcgTWFya3VwKHZhbHVlT3JTdHJpbmdzKTtcbiAgfVxuICBjb25zdCBzdHJpbmdzID0gdmFsdWVPclN0cmluZ3M7XG4gIGxldCBhY2MgPSBcIlwiO1xuICBsZXQgaSA9IDA7XG4gIGZvciAoOyBpIDwgcGxhY2Vob2xkZXJzLmxlbmd0aDsgKytpKSB7XG4gICAgYWNjICs9IHN0cmluZ3NbaV0gKyBodG1sRXNjYXBlKHBsYWNlaG9sZGVyc1tpXSk7XG4gIH1cbiAgYWNjICs9IHN0cmluZ3NbaV07XG4gIHJldHVybiBuZXcgTWFya3VwKGFjYyk7XG59XG5mdW5jdGlvbiBtYXJrdXAyKHN0cmluZ3MsIC4uLnZhbHVlcykge1xuICByZXR1cm4gc3RyaW5ncy5yZWR1Y2UoKHJlcywgc3RyLCBpKSA9PiByZXMgKyBzdHIgKyBodG1sRXNjYXBlKHZhbHVlc1tpXSkpXG59XG5yZXR1cm4ge01hcmt1cCwgbWFya3VwLCBtYXJrdXAyLCBodG1sRXNjYXBlfSJ9fQ)

LucasLefevre: Hi :wave: 

The reduce version is missing the accumulator initial value, the two versions are not doing the same thing. Once corrected, it appears the for-loop version is ever so slightly faster. But micro-benchmarks usually can't really be trusted to reflect real world scenarios anyway (engine optimization, etc.)

image

seb-odoo: Thank you for checking it! I'm pretty sure I compared both results to make sure they were the same, but I guess I must have missed something (or maybe I tried to "improve" the code after doing the check and didn't check again).

I get similar results as you once corrected. The for loop seems to be 3x faster for trivially small input (no param), and both versions become almost identical as the number of params increase.

sdegueldre: We generally try to avoid reduce for anything non-trivial (doing a sum or product) because of readability concerns, in this case the original version is less readable because it handles the fact that the placeholders are 1 item shorter than the strings array which the reduce version fails to handle and ends up appending "undefined" at the end.

I don't think performance is much of a consideration here, the strings being interpolated will always be few, even 10 interpolated pieces in a single markup literal would be a huge amount, but even then: on large string/placeholder counts the performance characteristics are dominated by _escapeHtml anyway (~70% of total time), not the concatenation. Using a single pass regex replaceAll with a replacer function speeds up the entire benchmark by a factor ~3. If we want to keep bike-shedding on performance that's where we should be looking but as I said I don't think it's particularly relevant. Markups are an escape hatch and are few and far between in the first place.

At this point, if this is a feature we want we should just merge it, if there are performance considerations that need to be addressed we can always circle back to this but I suspect this is just never going to be anywhere close to the bottleneck.

sdegueldre (approved):

seb-odoo:
Thank you for accepting it, it is a nice improvement!
This will remove the need of overriding the ci/security in many cases by having safer code by default.

I still wish `htmlEscape` was exported from here rather than having to re-define exactly the same function in Odoo though. I guess I can make a PR for it now that this one is merged.

ged-odoo:
we need to do something about markup and eventbus at least. Ideally, it should be accessible in some parts of odoo without the need for owl, for example, the website. there is a task for that though https://www.odoo.com/odoo/project/133/tasks/4364029?debug=assets

seb-odoo:
Should start re-exporting `markup` in `html.js` utils (from odoo) and use this import in business code rather than the one from owl then?

seb-odoo:
Export of the escape function + tests: https://github.com/odoo/owl/pull/1677

I still think it has value until a proper solution to the above-mentioned task is found.

Feel free to discard the PR, as we already have this function inside Odoo, it's not really "blocking".

seb-odoo:
And a final request on this matter, would it be possible to make a release with the current PR (and the escape one, if it is approved) one of these days? 🙏

I'd love to start using the new syntax in Odoo whenever possible! We do have quite a few markup-sensitive PR in the way (especially the composer one), and I'd like to rebase and finish https://github.com/odoo/odoo/pull/199300 too.

PR #1676, opened 2025-03-25
[FIX] utils: Correct validation of mount target in shadow DOM/iframe

abz-odoo:
Previously, `validateTarget` only checked if the target element or its host (if it was a ShadowRoot) was directly contained in the document body. This failed in cases where the target element was nested inside a shadow DOM, which itself was attached to the document.

This commit introduces a new helper `isAttachedToDocument` that traverses through parent nodes and shadow roots to ensure that the target is ultimately attached to the given document. Additionally, it now throws a clear error if `document.defaultView` is missing, indicating that the target document is detached or invalid.

This ensures proper validation of mount targets, including complex scenarios with shadow roots and iframes.

Before this commit, this threw an OwlError
```js
const shadow = fixture.attachShadow({ mode: "open" });
const shadowDiv = document.createElement("div");
shadow.append(shadowDiv);
const app = new App(SomeComponent);
const comp = await app.mount(shadowDiv);
```

Now this is a valide case

### DIFF
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -35,13 +35,43 @@ export function inOwnerDocument(el?: HTMLElement) {
   return rootNode instanceof ShadowRoot && el.ownerDocument.contains(rootNode.host);
 }
 
+/**
+ * Determine whether the given element is contained in a specific root documnet:
+ * either directly or with a shadow root in between or in an iframe.
+ */
+function isAttachedToDocument(
+  element: HTMLElement | ShadowRoot,
+  documentElement: Document
+): boolean {
+  let current: Node = element;
+  const shadowRoot = documentElement.defaultView!.ShadowRoot;
+  while (current) {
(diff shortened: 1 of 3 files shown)


src/runtime/utils.ts
```diff
@@ -35,13 +35,42 @@ export function inOwnerDocument(el?: HTMLElement) {
...
+  documentElement: Document
+): boolean {
+  let current: Node = element;
+  while (current) {
+    if (current === documentElement) {
+      return true;
+    }
+    if (current.parentNode) {
+      current = current.parentNode;
+    } else if (current instanceof ShadowRoot && current.host) {
```
ged-odoo, line 53: i guess that technically, you should check if it is an instance of document.defaultView.ShadowRoot. Otherwise, the instanceof check will not work cross iframe

ged-odoo (approved):

PR #1677, opened 2025-03-25
[IMP] runtime/utils: export htmlEscape and add tests

seb-odoo:
markup tag function requires markup awareness to determine whether a given parameter should be escaped or not.

This implies that pre-escaped content should be properly marked'ed up to avoid double escaping. Having to manually wrap all calls to escape with markup is cumbersome and prone to issues (on top of having to be validated by the security team for no reason).

This commit introduces a markup-aware escape function to resolve those issues.

### DIFF
--- a/tests/utils.test.ts
+++ b/tests/utils.test.ts
@@ -1,4 +1,4 @@
-import { batched, EventBus, markup } from "../src/runtime/utils";
+import { batched, EventBus, htmlEscape, markup } from "../src/runtime/utils";
 import { nextMicroTick } from "./helpers";
 
 describe("event bus behaviour", () => {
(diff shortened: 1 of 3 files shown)


tests/utils.test.ts
```diff
@@ -78,6 +78,56 @@ describe("markup", () => {
...
+    test("htmlEscape produces string from number", () => {
+      const res = htmlEscape(10);
+      expect(res.toString()).toBe("10");
+      expect(res).toBeInstanceOf(Markup);
+    });
+    test("htmlEscape produces string from boolean", () => {
+      const res = htmlEscape(false);
+      expect(res.toString()).toBe("false");
+      expect(res).toBeInstanceOf(Markup);
+    });
```
seb-odoo, line 106: I just copied the current implementation but I think these spec are far from ideal because they make "empty" values actually not empty.
The server might return false for a field, we don't want to display "false", but nothing instead.
We should probably just pass-through all empty values instead.

ged-odoo:
hmm, it looks like you want to export the htmlescape function to use it out of owl. in that case, you forgot to expose it as an exported value. it should also be exported from here https://github.com/odoo/owl/blob/master/src/runtime/index.ts

ged-odoo (approved):

PR #1694, opened 2025-09-03
[DOC] `slots` prop validation

bouvyd:
Make it explicit that props validation should accept `slots` if a component uses slots (even the `default` slot).


ged-odoo (approved):

ged-odoo:
@bouvyd agreed, but the formatting check fails

bouvyd:
Blazing fast 🚀 

Thx!

PR #1698, opened 2025-10-01
[FIX] error_handling: onError works if handling happens between culpr…

kebeclibre:
…it and root

Have a component implementing onError that calls a callback from one of its parent. That parent should not be the original source of the rendering (a parent above is).

Make the callback handle the error and trigger an render()

Before this commit, the parent did not see it was handling a subtree in error, and did not have a chance to revert that error state in its rendering stack.

After this commit, this flow works.

### DIFF
--- a/src/runtime/error_handling.ts
+++ b/src/runtime/error_handling.ts
@@ -55,6 +55,7 @@ export function handleError(params: ErrorParams) {
     let current: Fiber | null = fiber;
     do {
       current.node.fiber = current;
+      fibersInError.set(current, error);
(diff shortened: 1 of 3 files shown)


kebeclibre:
@ged-odoo 
This will help fix the use case in Odoo's navbar we spotted a few weeks back
(have systray item in error ; the wrapping in the Component ErrorHandler).

Note that I can fix the navbar/systray case in Odoo's space. What do you think ?

ged-odoo (approved):

PR #1859, opened 2026-04-21
Master misc2 ged

ged-odoo:


doc/reference/portal.md
```diff
@@ -0,0 +1,140 @@
+# Portal
+
+`<Portal>` is a built-in component that renders its slot content into a
+target DOM element somewhere else on the page, instead of in its own
+location in the parent's tree. It is the standard tool for modals,
+tooltips, popovers, dropdowns, toasts, and anything else that needs to
+escape its parent's overflow/stacking context.
```
sdegueldre, line 7: Are we changing the overlay mechanism in standard Odoo to use portals?

ged-odoo: in odoo, probably not. but yeah, the documentation here is partially auto generated, i will do a lot of documentation work later, when the code will stabilize

PR #1873, opened 2026-04-24
Master owl3 beta - EXPERIMENTAL - DO NOT MERGE

ged-odoo:


packages/create-owl/README.md
```diff
@@ -0,0 +1,33 @@
...
+Omit the project name to be prompted for one.
+
+## What you get
+
+A minimal, working Owl project with:
+
+- **Vite** dev server with HMR
+- **TypeScript** (or JavaScript, if you prefer)
+- A root component wired up and mounted
+- `npm run dev` / `npm run build` / `npm run preview`
```
sdegueldre, line 28: Would be cool if there was an option to choose ahead-of-time template compilation and to only ship the runtime so this is viable for browser extensions.

ged-odoo: oh wow, didn't think of that, but that's a pretty good idea. i'll add it to my todo list

sdegueldre: This will likely require re-exposing compile_templates as a "bin" (runnable command) at the package.json level

ged-odoo: will be a good opportunity to have a good look at it, and try to make it a little bit more usable. but this is not a high priority, this branch is just a dump of various ideas that we will probably, but not for sure, merge into odoo/owl in the future

PR #1922, opened 2026-06-05
Master ref signal ged

ged-odoo:


packages/owl-core/src/types.ts
```diff
@@ -308,7 +308,16 @@ function reactiveValueType(type?: any): ReactiveValue<any> {
...
 function ref<T extends Constructor<HTMLElement>>(type: T): InstanceType<T> | null;
 function ref(type?: any): any {
-  return union([literalType(null), instanceType(type)]);
+  // when no constructor is given, resolve HTMLElement lazily (at validation
+  // time) so that creating the schema works in non-DOM environments
+  const instance = type
+    ? instanceType(type)
+    : function validateHTMLElement(context: ValidationContext) {
+        if (typeof HTMLElement === "undefined" || !(context.value instanceof HTMLElement)) {
+          context.addIssue({ message: "value is not an instance of 'HTMLElement'" });
```
mcm-odoo: Do we want to fail if we are in a non-DOM env?
Also I don't believe that HTMLElement would start existing by itself. Either it is already existing at file load nor never exists.

```js
if (typeof HTMLElement === "undefined") {
  return t.any(); // or always failing validator
}
return union([literalType(null), instanceType(type || HTMLElement)]);
```

PR #1931, opened 2026-06-09
[IMP] doc: add more explanation on typing signals

### LINKED ISSUE #1923: allow to specify the return value of signal (in types of props)
props allow proper types/autocomplete to be defined for the most part, which is great.

But for signal, only the generic type seems allowed (`ReactiveValue<any, any>`) when calling `types.signal()`.

It would be nice to be able to define the return type of the signal as part of the types definition. For validation if possible, but mainly for autocomplete.

It's possible to do it manually with a second line, but it is not as clean as it requires a second line and repeating the signal word with a different syntax.
```js
            /** @type {import("@odoo/owl").Signal<string>} */
            value: types.signal(),
```

Note: there is a similar issue when creating a signal manually to a lesser extend. It can infer types properly when a value is given explicitly to the function, but not when the signal is initially empty, which typically happens for signals passed to `t-ref`. It would be nice to be able to "type" them on a single line (or with a special instruction) rather than with `/** @type {import("@odoo/owl").Signal<Element>} */` (or subclass of `Element`), and maybe the signal could reject values that don't match its type?

ged-odoo:
closes #1923


doc/v3/owl/reference/signals.md
```diff
@@ -0,0 +1,130 @@
...
+## Typing signals
+
+The type of a signal is normally inferred from its initial value, so `signal(0)`
+is a `Signal<number>`. When the initial value does not pin the type down — an
+empty array, or `null` — `signal` accepts a second `options` argument with a
+`type` to specify it explicitly:
+
+```js
+const ids = signal([], { type: [] as number[] }); // Signal<number[]>, not Signal<never[]>
+const selected = signal(null, { type: null as User | null }); // Signal<User | null>
```
mcm-odoo: Types are given in typescript instead of using validators
```js
const ids = signal([], { type: t.number() }); // Signal<number[]>, not Signal<never[]>
const selected = signal(null, { type: t.or([t.instanceOf(User), t.literal(null)]) }); // Signal<User | null>
```

PR #1938, opened 2026-06-11
Master optional ged

ged-odoo:


LucasLefevre:
Coucou @ged-odoo :wave: 

Personal feeling and feel free to ignore it :wink: 

`.default(value)` was easier to understand than `.optional(value)` to me.

What does `.optional(false)` mean ?
Does it mean it's *not* optional because I set it to `false` ?

While `.default(false)` or something like `.optional({ default: false })` are immediately clear.

Just a question of habit and, once you know it, you know it, but not super clear the first time reading

PR #1948, opened 2026-06-18
[FIX] playground: handle unicode in shared snapshots

### LINKED ISSUE #1946: Failed to share code (InvalidCharacterError)
1. Open the playground : https://odoo.github.io/owl/playground/#getting_started
2. Copy paste the content into `main.js`
```js 
import { Component, mount, signal, props, xml, types as t, toRaw, applyDefaults } from "@odoo/owl";

const propType = t.and([
    t.object({
        label: t.string().optional(),
    }),
    t.object({
        label: t.string().optional(),
    }),
]);

class Counter extends Component {
    static template = xml`
        <button t-on-click="this.increment">
          <t t-out="this.properties.label"/>: <t t-out="this.value()"/>
          <br/>
          Title : <t t-out="this.properties.title"/>
        </button>`;

    // note that the prop is fixed: it cannot change!
    properties = applyDefaults(props.static("properties", propType), propType);
    value = signal(0);
    
    increment() {
        this.value.set(this.value() + 1);
    }

    setup() {
      console.log(this.properties);
    }
}

class Root extends Component {
    static components = { Counter };
    static template = xml`
        <Counter properties="{label: 'Apples'}"/>
        <Counter properties="{label: 'Bananas', title: 'Minions'}"/>`;
}

mount(Root, document.body, { templates: TEMPLATES, dev: true });
```
3. Press the share button

The following error will be displayed in the console : 
```
playground.js:4376 Uncaught InvalidCharacterError: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
    at Playground.shareCode (playground.js:4376:19)
    at callHandler (owl.js:3779:6)
    at hdlr_fn3 (eval at compile (owl.js:6655:12), <anonymous>:8:33)
    at Object.mainEventHandler (owl.js:4224:7)
    at HTMLAnchorElement.listener (owl.js:1932:13)
shareCode @ playground.js:4376
callHandler @ owl.js:3779
hdlr_fn3 @ VM8:8
mainEventHandler @ owl.js:4224
listener @ owl.js:1932
```

piso-odoo:
The share feature encodes snapshots with `btoa()`, which fails when the content contains non-Latin1 characters.

Encode snapshots as UTF-8 before converting them to base64.

closes #1946

### DIFF
--- a/tools/playground/src/playground.js
+++ b/tools/playground/src/playground.js
@@ -106,7 +106,8 @@ class Playground extends Component {
         this.hashTutorial = tutorial;
       } else {
         try {
-          const parsed = JSON.parse(atob(hash));
+          const decodedHash = decodeURIComponent(atob(hash));
+          const parsed = JSON.parse(decodedHash);
           if ("js" in parsed && typeof parsed.js === "string") {
             const { js, css, xml } = parsed;
             const fileNames = [];
@@ -239,7 +240,8 @@ class Playground extends Component {
 
   shareCode() {
     const snapshot = this.code.getSnapshot();
-    const state = btoa(JSON.stringify(snapshot));
+    const encodedSnapshot = encodeURIComponent(JSON.stringify(snapshot));
+    const state = btoa(encodedSnapshot);
     const link = new URL(window.location.href);
     link.hash = state;
     link.searchParams.set('v', this.owlVersion.version());


sdegueldre:
Doesn't this break existing shared playground links?

piso-odoo:
No, as for decoding I am just adding a extra `decodeURIComponent` and  this will just return same string if it is a valid UTF-8 which should be the case in existing playground links.

ged-odoo (approved):

PR #1956, opened 2026-06-23
[IMP] playground: enhance playground with Monaco editor with OWL tooling

piso-odoo:
This PR improves the OWL Playground editing experience after the migration to Monaco Editor by adding richer language features, editor productivity enhancements, etc.

### Playground
- Replace CodeMirror with Monaco Editor.
- Add version-aware Owl TypeScript declarations based on the selected Owl version.
- Support automatic imports:
  - Auto-detect and import symbols from existing project files.
  - Automatically provide `@odoo/owl` imports in the playground.

### Tooling
- Generate Shiki grammar files during build from source grammars.
- Share the grammar transformation pipeline between documentation and playground.
- Remove duplicated Shiki grammar definitions.


tools/playground/libs/workers/ts.worker.js
ged-odoo, line 1: **These generated bundles shouldn't be committed — gitignore them and build in CI**

This file (`ts.worker.js`, ~7 MB) plus the rest of `libs/workers/*` and `libs/monaco/*` are ~21.6 MB of minified output generated by `build_monaco.mjs`. They're currently load-bearing: neither `build:site` nor `pages.yml` runs `build:monaco`, so the deploy relies on the committed copies — which means every `monaco-editor`/`shiki` bump re-commits ~20 MB into git history.

Since they're fully reproducible from devDependencies, I'd suggest treating them like the existing build output (`dist/` is already gitignored):

**1. Gitignore the generated files:**
```gitignore
/tools/playground/libs/monaco/
/tools/playground/libs/workers/*.worker.js
/tools/playground/static/monaco.bundle.css
/tools/playground/static/codicon.ttf
```
The `*.worker.js` glob ignores the four generated workers while keeping the hand-written `custom_ts_worker.js` tracked. `monaco.bundle.css` and `codicon.ttf` are also emitted into `static/` by the same script, so they get the same treatment.

**2. Wire `build:monaco` into the pipeline before `assemble.mjs`:**

Root `package.json` → `build:site`:
```diff
-"build:site": "npm run build && npm run build:playground && npm run build:doc && npm run build:doc-v2 && node tools/site/assemble.mjs",
+"build:site": "npm run build && npm run build:playground && npm run build:monaco -w tools/playground && npm run build:doc && npm run build:doc-v2 && node tools/site/assemble.mjs",
```

`.github/workflows/pages.yml`, add a step before the `assemble.mjs` run:
```yaml
- run: npm run build:monaco -w tools/playground
```

Cost is one build step + pulling `monaco-editor`/`shiki` at install (already devDependencies). In exchange the repo stays lean and Monaco upgrades become a lockfile bump instead of a 20 MB diff.

piso-odoo: I have added that in build step.

PR #1964, opened 2026-06-25
[FIX] playground: adapt `.optional()` instead of `?`

vpan-odoo:
In this commit:
===============
- Adapt new optional props and types validation `.optional()` instead of deprecated `?`


vpan-odoo:
@ged-odoo small PR fixing optional props in playground sample code. check it when you have time.

Thanks

ged-odoo:
@vpan-odoo thanks!

PR #1979, opened 2026-07-07
[FIX] runtime: don't lose coalesced renders

seb-odoo:
When a reactive change notifies a component, its reactive subscriptions are cleared before its render callback runs. If the component's parent holds a pending, not yet rendered fiber for it (the parent is mid-render, e.g. awaiting `onWillUpdateProps`), the requested render is coalesced into that pending fiber. If another parent render then cancels that fiber before it could render, `cancelFibers` only set `forceNextRender` for fibers that had already rendered (with a `bdom`): a not yet rendered fiber was dropped without compensation. When the child's props also compare equal to its committed props (a transient prop that changed and reverted compares equal, as committed props are only updated on patch), the cancelling render pass skips the child entirely: the coalesced render is lost, and since the subscriptions were already cleared, the component no longer reacts to any further change of the reactive state it displays. Its DOM is silently stale until an unrelated render of it happens.

Force the next render of a mounted node whose not yet rendered fiber is cancelled, as master already does.

This is the root cause of odoo runbot error 940032 ("thread messages never render"): the render displaying the freshly loaded messages of a discuss channel was coalesced into a pending fiber that a parent double-render cancelled, and the message list stayed empty with no error, surviving two application-side fix attempts.

https://runbot.odoo.com/odoo/error/940032

### DIFF
--- a/src/runtime/fibers.ts
+++ b/src/runtime/fibers.ts
@@ -76,6 +76,15 @@ function cancelFibers(fibers: Fiber[]): number {
       node.forceNextRender = true;
     } else {
       result++;
+      if (node.bdom) {
+        // a mounted node whose pending fiber is cancelled before it could
+        // render may have renders coalesced into that fiber (requested by
+        // reactive changes, whose subscriptions were cleared when they were
+        // notified), so dropping the fiber would silently lose them: force
+        // the cancelling render to re-render the node even if its props are
+        // unchanged.
+        node.forceNextRender = true;
+      }
     }
     result += cancelFibers(fiber.children);
   }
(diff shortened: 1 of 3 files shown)


seb-odoo:
@ged-odoo here's a proposed OWL 2 fix

Disclaimer: AI helped me here, I'm not fully sure to grasp everything. It at least made sense to me, but maybe you would have a better solution.

Similar "business" test in Odoo (slightly technical as it's a timing issue): https://github.com/odoo/odoo/pull/274703

ged-odoo:
@seb-odoo it looks like a good fix. see this commit in owl 3 for comparison: https://github.com/odoo/owl/commit/cbe1deaf1c829786d27bf48c4469de488ae7c277

seb-odoo:
Indeed the fix looks the same. Should I try to back-port the same "comment" and test, or are we ok with slight differences?

ged-odoo:
no it's fine i think. owl 2 is not really in a very active state anyway

PR #1987, opened 2026-07-08
[IMP] playground: initialize Monaco only once

piso-odoo:
Switching between the editor and project manager remounts the CodeEditor, causing `onWillStart` to re-register Monaco workers, snippets, themes, and language configuration on every mount.

Move Monaco initialization into a singleton helper so the global setup runs only once per page.

This avoids duplicate Monaco initialization when reopening the editor.

### DIFF
--- a/tools/playground/src/monaco/initialize_monaco.js
+++ b/tools/playground/src/monaco/initialize_monaco.js
@@ -1,0 +1,65 @@
+import { registerXmlTagRename } from "./xml_tag_rename.js";
+import { registerOwlSnippets } from "./snippets.js";
+import { setupShiki } from "./shiki.js";
+import { registerCustomTsWorker } from "./auto_import.js";
+import { playgroundAssetUrl } from "../asset_url.js";
+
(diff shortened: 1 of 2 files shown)


tools/playground/src/monaco/initialize_monaco.js
```diff
@@ -0,0 +1,65 @@
...
+    });
+    monaco.typescript.javascriptDefaults.addExtraLib(
+        `declare const TEMPLATES: Record<string, string>;`,
+        "file:///globals.d.ts"
+    );
+    registerOwlSnippets(monaco);
+    registerXmlTagRename(monaco);
+    await registerCustomTsWorker(monaco);
+    await setupShiki(monaco);
+    monacoInitialized = true;
```
ged-odoo: this should be done above. at line 13, to avoid a situation where we still quickly manage to initialize monaco twice

piso-odoo: Moved up that line.

PR #1995, opened 2026-07-14
[FIX] owl-core: update asyncComputed state after disposal

### LINKED ISSUE #1994: Fix asyncComputed state updates after disposal
`asyncComputed` can still apply the result of a pending async run after the owning scope has been disposed. This can lead to stale values or errors being written back to the computed state even though the computation is no longer active. The helper should ignore late results once it has been disposed.

When an `asyncComputed` is created inside a component/scope and the scope is destroyed while the fetch is still pending, a later resolve/reject can still call `value.set(...)`, `error.set(...)`, or otherwise update the reactive state even though the computation is no longer valid.

For Example: 

- a component creates an `asyncComputed` to fetch data,
- the component is destroyed or unmounted before the request finishes,
- the promise later resolves,
- without a disposal guard, the result still updates the computed state even though the component is gone.

sauc-odoo:
When an `asyncComputed` is created inside a component/scope and the scope is destroyed while the fetch is still pending, a later resolve/reject can still call `value.set(...)`, `error.set(...)`, or otherwise update the reactive state even though the computation is no longer valid.

After this commit, once an `asyncComputed` is disposed, late results from it should be ignored.

fixes https://github.com/odoo/owl/issues/1994

### DIFF
--- a/packages/owl-core/src/async_computed.ts
+++ b/packages/owl-core/src/async_computed.ts
@@ -128,6 +128,7 @@ export function asyncComputed<T>(
   });
 
   function dispose() {
+    runId++;
     stopEffect();
     runController?.abort();
     runController = null;
--- a/packages/owl-runtime/tests/reactivity/async_computed.test.ts
+++ b/packages/owl-runtime/tests/reactivity/async_computed.test.ts
@@ -468,6 +468,31 @@ test("currentPromise() resolves on dispose so awaiters do not hang", async () =>
   expect(resolved).toBe(true);
 });
 
+test("dispose prevents a late resolution from mutating the asyncComputed state", async () => {
+  const def = makeDeferred<number>();
+  const a = asyncComputed(() => def, { initial: 0 });
+
+  a.dispose();
+  def.resolve(42);
+  await flush();
+
+  expect(a()).toBe(0);
+  expect(a.error()).toBeNull();
+});
+
+test("dispose prevents a late rejection from mutating the asyncComputed state", async () => {
+    const def = makeDeferred<number>();
+    const a = asyncComputed(() => def, { initial: 0 });
+
+    a.dispose();
+
+    def.reject(new Error("boom"));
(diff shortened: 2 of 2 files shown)


packages/owl-runtime/tests/reactivity/async_computed.test.ts
```diff
@@ -468,6 +468,22 @@ test("currentPromise() resolves on dispose so awaiters do not hang", async () =>
...
+
+  a.dispose();
+  def.resolve(42);
+  await flush();
+
+  expect(a()).toBe(0);
+  expect(a.loading()).toBe(false);
+  expect(a.error()).toBeNull();
+});
+
```
ged-odoo, line 482: if you can, add a test to make sure that if we reject the deferred with an error after dispose, it behave as expected (a.error() stays null)

packages/owl-core/src/async_computed.ts
```diff
@@ -49,6 +49,7 @@ export function asyncComputed<T>(
 
   let runId = 0;
   let runController: AbortController | null = null;
+  let disposed = false;
```
ged-odoo: it feels like disposed is doing basically the same job as runId. instead of using a boolean variable here, we can just change runId (increment it, or set it to -1) to have the same effect as

ged-odoo (commented):
excellent

packages/owl-core/src/async_computed.ts
sauc-odoo: Hello @ged-odoo 
made the changes, please check it. Thanks!!

packages/owl-runtime/tests/reactivity/async_computed.test.ts
sauc-odoo: added

ged-odoo (approved):
code is simpler at the end. it's a good sign :)

PR #2004, opened 2026-07-25
[FIX] owl-runtime: don't notify prop signals for alike props

### LINKED ISSUE #2003: investigate: avoid updating arrow functions props if possible
## Context:

> Props assigned an arrow function are "ignored" (auto .alike) for re-renders 👍
> ... but they are not ignored for useEffect if a render actually occurs for another reason 🤔
> 
> Another way to say it: I think the intended behavior of useEffect is to "trigger when something that is read inside the effect changes" but a simple useEffect(() => { this.props.doStuff(); }) actually triggers on all renders, as doStuff gets reassigned to a new arrow function. Is this wanted?
> 
> I made a PR with a quick test to demo the issue more precisely: https://github.com/odoo/odoo/pull/278104 (the last assertion fails, I would have imagined it should not).

> well, yes it is intended. the useEffect has no way of knowing that the function is actually the same. if the effect reads a reactive value (in this case this.props.doStuff), and that value is reassigned, the effect is reexecuted. Now, for the props in a template, since owl knows that the new function is actually the same closure, it does not need to update the child component.
> 
>  for the props in a template, since owl knows that the new function is actually the same closure, it does not need to update the child component

> If it didn't update that prop, wouldn't the problem go away too?

 
> that's an interesting thought. not sure how easy it is to propagate the information to the useProps function and define a signal with a truthy comparison function to skip the update, but it could very well work. i'll investigate

Gxrvish:
Props returned by useProps are signals, so reading one inside an effect subscribes to it. Until now, every prop signal was updated on each props update, even for the props the template does not compare by identity: `.bind` and `.alike` props, and inline arrow functions.

Those props are a new function object on every render, but they mean the same thing, so Owl already ignores them when it decides whether to re-render a child. Reactivity did not. As soon as the child was re-rendered for any other reason, its props object was replaced, the new function went into the signal, and every effect reading that prop ran again.

The compiler now passes the list of those props to createComponent, which stores it on the component node, and useProps skips the signal update when nothing the arrow function captured has changed. A prop signal now notifies in the same cases where Owl decides to re-render.

Note: a `.bind` prop now keeps the function it already had instead of picking up a newly bound one when the child re-renders for another reason. This is what `.alike` already means for rendering.

The extra argument is left out when a component has no such prop, so templates compiled with an older compiler behave as before.

Fixes: https://github.com/odoo/owl/issues/2003

### DIFF
--- a/doc/v3/owl/reference/props.md
+++ b/doc/v3/owl/reference/props.md
@@ -377,6 +377,29 @@ wrong to use `.alike`.
 </t>
 ```
 
+### Alike props and effects
+
+Props coming from `useProps` are reactive, so reading `this.props.someProp` in an
+[effect](effects.md) or a [computed value](computed_values.md) subscribes to it.
+Alike props are left out of that too, and do not notify when the parent renders
+again with a new function.
+
+```js
+class Child extends Component {
+  props = useProps(["value", "onClick"]);
+  setup() {
+    // runs again when `value` changes, but not because the parent rebuilt `onClick`
+    useEffect(() => {
+      console.log(this.props.value, this.props.onClick);
+    });
+  }
+}
+```
+
+An anonymous function still notifies if one of the values it captures changes,
+since that is also what makes Owl re-render the child. A `.alike` or `.bind` prop
+captures nothing, so it never notifies.
+
 ## Binding function props
 
 It is common to have the need to pass a callback as a prop. Since Owl components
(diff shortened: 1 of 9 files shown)


ged-odoo:
hey @Gxrvish i noticed your PR, thanks! I am still thinking about it, not sure if the cost outweigh the value. The useProps function is a function that is called many times, and i want to be careful. but i did not forget this, still not sure, that's it!

seb-odoo:
Shouldn't the real fix be to use static props for those never changing props?

I might be a little extreme but I'm thinking all props should always be static eventually.

And if their value must change, the static prop should contain a signal/computed to change it through reactivity, not through props change.

I guess the "optimization" might not hurt in the meantime.

PR #2014, opened 2026-08-21
[FIX] owl-runtime: make sub-root renders yield to ancestor renders

ged-odoo:
## Problem

With a Portal whose slot both reads a signal and is guarded by it in the owner:

```xml
<Popover t-if="this.isPopoverOpen()">
  Coucou <t t-if="console.log(this.isPopoverOpen())"/>
</Popover>
```

closing the popover logs `false` — the slot is evaluated with the very value that is removing it (a "zombie render"), even though it sits behind a `t-if` guarding against that value.

Slot content rendered through a Portal subscribes the portal **sub-root**'s computation, not the slot owner's. Flipping the signal schedules two re-renders: the owner's (whose `t-if` will remove the Portal) and the sub-root's. The sub-root rendered first.

Owl already prevents this for in-tree components: `Fiber.render` walks up `node.parent` looking for an in-flight ancestor render, and delays the child's render if the ancestor's pass dropped it (`delayedRenders`), so the pending render dies with the node at commit. But `createRoot` builds sub-root nodes with `parent = null`, so Portal (and Suspense) content escaped the walk entirely.

## Fix

Sub-roots now carry a `host` link (internal `SubRootConfig.host`, stored on `ComponentNode`, passed by Portal and Suspense) — not a parent link, just visibility for the scheduler. The ancestor walk climbs parent-or-host, with two boundary rules:

- The childrenMap retention test cannot apply at the boundary (sub-root content never appears in the host's childrenMap): crossing counts as retained when the host itself holds a fiber of a finished (counter 0) pass — a node holding a fiber of a finished pass survives it — and otherwise the walk continues into the host's ancestors, where the normal logic decides.
- The crossing only applies once the sub-root is **mounted** (`above()` helper). Suspense deliberately renders its default slot in parallel with the outer tree's still-in-flight mount, so the initial render must not be delayed; the zombie scenario is by definition a re-render of mounted content. (A first version without this gate broke three Suspense tests — fallback-flash and parallel-willStart behavior.)

## Tests

New regression test in `portal.test.ts` probing every evaluation of a portaled slot guarded by a `t-if`: it recorded `[true, false]` before the fix, `[true]` after — the slot is never evaluated with the guard value that removed it, and the portal target is emptied correctly. Full owl-runtime suite passes (1402 tests, Suspense suite included) plus the umbrella package.

### DIFF
--- a/packages/owl-runtime/src/app.ts
+++ b/packages/owl-runtime/src/app.ts
@@ -37,6 +37,9 @@ interface SubRootConfig<P> extends RootConfig<P> {
   // error handler seeded on the root node so descendant errors route back into
   // the host's parent chain instead of tearing down the whole app.
   onError?: (error: any, finalize: Function) => void;
+  // the component node hosting the sub-root, so the scheduler can delay
+  // renders inside the sub-root while an ancestor of the host is rendering.
+  host?: ComponentNode;
 }
 
 export interface AppConfig extends TemplateSetConfig {
@@ -139,6 +142,9 @@ export class App extends TemplateSet {
       }
       if (subConfig.onError) {
         nodeErrorHandlers.set(node, [subConfig.onError]);
+      }
+      if (subConfig.host) {
+        node.host = subConfig.host;
       }
     } catch (e) {
       error = e;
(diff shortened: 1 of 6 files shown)


ged-odoo:
Follow-up commit: the host link is no longer a field on ComponentNode — it now lives in a module-level `subRootHosts` WeakMap (same pattern as `nodeErrorHandlers`), so regular nodes pay nothing for this Portal/Suspense-only metadata.

aab-odoo (approved):

seb-odoo:
This is great news. Hopefully we can now remove all `t-if` we have in slots and stop consider them as dangerous to use!

PR #2016, opened 2026-08-21
[IMP] owl-core: add types for Set and Map

### LINKED ISSUE #1939: add types for Set and Map
Unless I missed how to do it, the only way to types Set and Map seems to use instanceof which doesn't allow to specify the types of the values (and keys for Map).

seb-odoo:
Before this commit, a set or a map could only be typed with `t.instanceOf(Set)`, which validates the container and nothing else: what it holds is unchecked, and the type derived from the schema is `Set<any>`. `t.customValidator` can check the content, but it keeps the type of the base it wraps, so the reader gets no completion.

This commit adds `t.set(valueType?)` and `t.map(keyType?, valueType?)`, which validate every element, key and value, and resolve to `Set<V>` and `Map<K, V>`. A failing entry is reported by its iteration index (`0 > key`, `0 > value` for a map), as a map key can be any value.

Note that the validation context gains `withEntry(key, value)`: a set element and a map key are not reachable by key on the validated value.

Closes odoo/owl#1939

### DIFF
--- a/doc/v3/owl/reference/types_validation.md
+++ b/doc/v3/owl/reference/types_validation.md
@@ -182,6 +182,35 @@ t.record(t.number()); // all values must be numbers
 // rejects:   { a: 1, b: "two" } with t.record(t.number())
 ```
 
+### `t.set(valueType?)`
+
+Validates that the value is a `Set`. When `valueType` is provided, every
+element is validated against it.
+
+```js
+t.set(); // any set
+t.set(t.number()); // set of numbers
+
+// validates: new Set([1, 2])     with t.set(t.number())
+// rejects:   new Set([1, "two"]) with t.set(t.number())
+```
+
+### `t.map(keyType?, valueType?)`
+
+Validates that the value is a `Map`. When `keyType` is provided, every key is
+validated against it, and likewise for `valueType` and the values. A failing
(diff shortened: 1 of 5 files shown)


ged-odoo:
i think you should do something like this: https://github.com/odoo/owl/commit/ad592dd61c88e23e8e5e59467d4659560fd00d03 to avoid observing values when validating types

seb-odoo:
@ged-odoo but then the issue is generic and pre-existing, it should apply to all types, including custom validators that use signal internally, so `untrack` should be used. `toRaw` is only useful for performance when walking through `proxy` (and kills the reactivity as a side effect, which would make it inconsistent with signal that are unaffected).

I pushed a new commit to take this into account, but I'm not sure if we really need this.

Is there anything actually observing while validate?

If the validator is called in an effect/computed like in the tests, wouldn't it be the expected behavior that it actually re-validates on change?

seb-odoo:
Should I push the PR without the no-observing commit, so we can already merge the Set/Map, and determine later/when we have time whether observing should be handled or not and how?

seb-odoo:
I rebased to fix issues. I also removed the no-observe commit so the Set/Map types can be merged on their own. The untrack/toRaw part is kept aside, we can decide later if validation should observe or not.

PR #2018, opened 2026-08-25
[FIX] owl-core, owl-runtime: handle String objects

seb-odoo:
Two fixes for `String` objects, which owl already accepts at runtime: a `Markup`, or a string subclass such as a lazily translated term.

The first commit types `markup(value)` and `.optional(default)` for them. The second one makes `VText`, `VHtml` and the compiled-block text slot compare the values as text when patching, so a `String` built again on every render no longer rewrites or re-parses unchanged content.

Details in each commit message.


seb-odoo:
Note: somewhat related Odoo PR: https://github.com/odoo/odoo/pull/284440

seb-odoo:
Rebased

PR #2028, opened 2026-08-31
[FIX] doc: correct v3 reference/playground inaccuracies found by doc audit

jpp-odoo:
## Summary

Fixes 15 confirmed inaccuracies found by a full-repo audit of `doc/v3` and `tools/playground/samples/v3` against current source (`packages/{owl-core,owl-compiler,owl-runtime,owl}`):

- **Component & App core** — `app.md`'s `Root` object claimed a `node` property that doesn't exist (real fields: `prepared`/`destroyed`) and a fabricated "3-second onWillStart deadlock warning"; `overview.md` pointed at the deprecated `config` alias instead of `useConfig`.
- **Hooks & Effects** — `effects.md` claimed an effect created inside a plugin is disposed with that plugin; it's actually bound to the shared plugin manager's lifetime.
- **Props & Bindings** — `form_bindings.md` didn't disclose that `.trim` silently forces `.lazy`-style change-event timing too.
- **Templates** — `template_syntax.md` showed an invalid `t-key` placement and a false "falls back to index" claim (it throws), plus a wrong custom-directive output example; `precompiling_templates.md` referenced a nonexistent `owl.iife.runtime.js` build artifact (v2 leftover); `translations.md`'s translatable-attribute list was missing 4 `aria-*` attributes.
- **DOM & Control Flow** — `error_boundary.md` pointed at a dead `t-ref`/`useRef` access path (`t-ref` isn't supported on components in v3).
- **Plugins & Infra** — `scope.md` documented a nonexistent `Scope.parent` field and examples importing the never-exported `STATUS` enum.
- **Migration & Design Narrative** — `owl3_design.md` claimed a read-only computed's `.set()` silently no-ops (it throws), and filed `useProps.static()` as a rejected idea despite it being shipped.
- **Playground samples** — `web_client/core/orm.js` imported the nonexistent `useContext` instead of `useScope`; `web_client/readme.md` misplaced a file in its documented structure; two tutorial readmes (`todo_list/7`, `hibou_os/12`) taught a `"key?"` optional-prop syntax that doesn't work (real syntax is `.optional()`, matching each step's own solution file).
- **Playground architecture doc** — `playground.md` was rewritten from scratch: it described a prior CodeMirror-based playground (wrong editor, self-contradicting plugin count, 6/9 fabricated sample IDs, a tutorial that was never built, a "Standalone Export" feature bundling files + a Python server that don't exist, wrong file structure and component line citations) and omitted Vim mode, dark mode, and the v2/v3 version switcher entirely. Rewritten against the current `tools/playground/src/` implementation.

## Test plan

- [ ] Doc-only + one sample file (`orm.js`) change — no build/test impact expected
- [ ] `npm run test:doc-links` (checks doc cross-references still resolve)
- [ ] Spot-check `playground.md` renders correctly in the playground's own markdown viewer

### DIFF
--- a/doc/v3/owl/reference/template_syntax.md
+++ b/doc/v3/owl/reference/template_syntax.md
@@ -468,8 +468,8 @@ The `t-key` directive is useful for lists (`t-foreach`). A key should be
 a unique number or string (objects will not work: they will be cast to the
 `"[object Object]"` string, which is obviously not unique).
 
-Also, the key can be set on a `t` tag or on its children. The following variations
-are all equivalent:
+Also, the key can be set on a `p` tag or on a wrapping `t` tag. The following
+variations are equivalent:
 
 ```xml
 <p t-foreach="this.items" t-as="item" t-key="item.id">
@@ -479,13 +479,11 @@ are all equivalent:
 <t t-foreach="this.items" t-as="item" t-key="item.id">
   <p t-out="item.text"/>
 </t>
-
-<t t-foreach="this.items" t-as="item">
-  <p t-key="item.id" t-out="item.text"/>
-</t>
 ```
 
-If there is no `t-key` directive, Owl will use the index as a default key.
+`t-key` must be set on the same element as `t-foreach`, not on one of its
+children. If it is missing, Owl throws `"Directive t-foreach should always
+be used with a t-key!"` — there is no fallback to the index.
 
 Note: the `t-foreach` directive only accepts arrays (lists) or objects. It does
 not work with other iterables, such as `Set`. However, it is only a matter of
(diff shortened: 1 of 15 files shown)


doc/v3/owl/reference/template_syntax.md
ged-odoo: should probably put this.click for this example

doc/v3/owl/reference/template_syntax.md
```diff
@@ -623,7 +621,7 @@ This :
 will be replaced by :
 
 ```xml
-<div t-on-click="value"/>
+<div t-on-click="click"/>
```
ged-odoo: and here

doc/v3/owl/reference/template_syntax.md
jpp-odoo: Fixed in e86fde10 — switched the example to `this.click`.

doc/v3/owl/reference/template_syntax.md
jpp-odoo: Fixed in e86fde10 — switched the example to `this.click`.

PR #2032, opened 2026-09-09
[IMP] owl: declare signal and static props in the schema

### LINKED ISSUE #1907: types validation for ReactiveValue and read-only passing
I have a component that takes indifferently a `signal` or a `computed` as props (returning a string). The only thing that matters is the `()` to read the value.

```js
            /** @type {import("@odoo/owl").ReactiveValue<string|undefined>} */
            value: types.signal(),
```

The type hint suggests that I should use `ReactiveValue` for a generic type , but this type seems missing from `types`.
Why is there even a `ReactiveValue` and a `Signal` interface if both pretty much have the same API, and only a minor difference in the `set` dogstring?

And inversely, why isn't there a specific type for `Computed` that do have a different API in regard to the existence of the set function and the input parameter of it.

Should I just use `signal` from `types` in that case?

And what if the component wants to clarify that it does not intend to write on the signal/computed?

Do we have or could we have a way to pass a signal or a computed as read-only enforced, and a way to type it?

### LINKED ISSUE #1909: optimize .signal= props passing
The documentation states that `.signal=` props passing is not a performance optimization, but I believe it could be.

In the case of `val.signal="expression"`.

Instead of executing the `expression` during rendering of the parent (which will trigger parent render on change to set the new value on the signal), the expression should instead be wrapped into a `computed(() => expression)` once, and `expression` will only be called in the context of the child.

The parent doesn't actually need to know nor care about what is inside the expression, it only needs to forward it and let the child do the observing as necessary for it.

I may be missing a case, but this is the most common occurrence of signal props passing that we need in our code, so it would be nice to optimize it. If indeed there is a missing case that makes this not possible to generalize, I would suggest to also introduce `.computed` props passing that has this behavior.

The use case that we have all the time is: we have a proxy, and we need to forward a non-signal field of the proxy as props, and we want it to be dynamically observed in child.

seb-odoo:
Before this commit, a signal prop cannot be declared in the schema: `useProps()` wraps a `t.signal` key like any other, so the child reads a signal of a signal, and declaring one takes `useProps.static(name, type)`, one call per prop, outside the schema. The schema also has no way to say that a component only reads the signal it receives, or that it writes on it, so handing a read-only value to a component that writes is only caught when the write crashes.

`.signal=` has its own limit: the expression is evaluated during the parent's render, so every change of what it reads re-renders the parent just to refresh the wrapper. The suffix adapts a plain value to a signal prop, it does not save a render.

These commits give the schema keys their own behavior and move the evaluation into the child:

- A `t.signal(...)` key is read once and exposed as is, read-only unless the prop is declared `settable`.
- A plain value on such a key is promoted through a signal owned by the child, so a child can declare `t.signal` without touching any caller.
- `.static()` brands any type, like `.optional()`, for a prop read once.
- A `.signal=` expression is compiled into an arrow evaluated inside the cached computed, so the parent's render never subscribes to what the expression reads and a change updates the child alone.

Read-only itself becomes the absence of `set`: a computed declared without a `set` option no longer carries one that throws.

Closes odoo/owl#1907
Closes odoo/owl#1909

### DIFF
--- a/doc/v3/owl/reference/props.md
+++ b/doc/v3/owl/reference/props.md
@@ -145,6 +145,62 @@ This only applies to props accessed through `useProps()`. The singular
 [`useProps.static()`](#the-usepropsstatic-method) helper keeps its static, reference-stable
 semantics.
 
+### Signal props
+
+A prop declared with `t.signal(...)` at the top level of the schema is a
+**signal prop**: the child reads it as a reactive value, and the prop itself
+is static (the parent must pass the same signal on every render; only its
+inner value changes).
+
+```js
+import { Component, useProps, t, xml } from "@odoo/owl";
+
+class Counter extends Component {
+  static template = xml`<span t-out="this.props.count()"/>`;
+  props = useProps({ count: t.signal(t.number()) });
+}
+```
+
+`this.props.count` is the reactive value itself, assigned once: reading it in
+the template, a computed or an effect subscribes to the signal, without any
+render of the parent involved.
+
+The child's view is **read-only by default**: when the parent passes a
+writable signal, the child receives a wrapping computed with no `set` method.
+A component that intends to write the received signal declares it:
+
+```js
+props = useProps({ value: t.signal(t.number(), { settable: true }) });
+```
+
+A settable prop exposes the original signal, and validation requires the
+received value to have a `set` method (a read-only computed is rejected).
+
+When the parent passes a **plain value** instead of a signal, the value is
+promoted: the child still reads a reactive value, updated when the parent
+re-renders with a new value. This keeps every caller working while they
+migrate to passing signals. A caller must not alternate between passing a
+plain value and a signal (dev mode throws).
+
+### Static props in the schema
+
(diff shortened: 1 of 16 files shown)


seb-odoo:
Hello @ged-odoo here is a tentative PR to fix what we believe are issues with the current props system.

I've adapted the current OWL branch into Odoo in these PR:

https://github.com/odoo/odoo/pull/287375
https://github.com/odoo/enterprise/pull/130936

This includes all breaking changes to make the branch green (there are not many), and some adaptations in our code to highlight the new syntax.

ged-odoo:
@seb-odoo thanks for this work. i am currently quite busy, but it looks interesting.

ged-odoo:
@seb-odoo 
- sorry, but there are now conflicts
- question: you say "The problem is that a read-only reactive
value cannot be told apart from a writable one". i wonder why that is a problem?

seb-odoo:
It's a problem if we want to do props validation, we need to be able to easily tell them apart without calling the method to see what it does.

I'll fix the conflicts.

PR #2035, opened 2026-09-10
[IMP] portal: add `ordered` to keep portals in declaration order

alexkuhn:
A Portal appends to its target, so portals sharing a target land in mount order. That is right for dialogs and popovers, but wrong when the target is an ordered list: a portal appearing later always lands last, and a keyed t-foreach reorder moves the portals without re-rendering, so nothing puts the target back in order.

A portal cannot fix this alone, knowing nothing of its siblings. `ordered="true"` joins it to a per-target group that sorts on the document position of the empty text node every Portal already leaves in place, and observes those nodes' parents to catch a reorder owl performs by moving nodes.

Needed by mail's `Tabs`, which portals its tab headers into the nav from a t-foreach.

https://github.com/odoo/odoo/pull/287612

### DIFF
--- a/packages/owl-runtime/src/portal.ts
+++ b/packages/owl-runtime/src/portal.ts
@@ -1,5 +1,6 @@
 import { Signal } from "@odoo/owl-core";
 import { Component } from "./component";
+import { ComponentNode } from "./component_node";
 import { useEffect } from "./hooks";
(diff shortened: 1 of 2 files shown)


alexkuhn:
Solved by https://github.com/odoo/owl/pull/2036

PR #2037, opened 2026-09-23
[FIX] owl-core: validate a value without observing it

seb-odoo:
Before this commit, validating inside a computation subscribes it to what it validates: an effect calling
`assertType(state.items, t.array(t.number()))` re-runs on every write to the array, and a `t.customValidator` predicate reading a signal subscribes to that signal.

This happens because the validators read the value through its proxy, which also allocates a per-key atom for every key they walk, even when no computation is active.

This commit runs `validateType` untracked and hands each validation context the raw value: nothing a validator or a custom predicate reads is observed, and a reactive value is walked without allocating anything. Note that an issue now reports the raw value in `received`.


seb-odoo:
Split from https://github.com/odoo/owl/pull/2016, if we actually want this feature.

But I'm not really sure. It feels like reactivity could be nice there, to validate "in real time".

I don't have a use case for or against this either way.

But in case of props validation: shouldn't we validate reactive props while their content change too, and not only when they are first passed?

seb-odoo:
So this is a proof of concept but I think my feeling would be to actually close this one.

